PR #9, opened 2019-12-16
adding detection example using stack, on simple_sim

esheldon:
- do a straight coadd and store it in an ngmix Observation for the simple_sim, which makes perfectly registered images with a pixel scale and
    same psf and wcs for all images
- run detection and deblending
- run a stub for measurement on deblended images (not doing anything yet)
- optionally make a plot and overplot detections on the image

### DIFF
--- a/examples/detect_simple.py
+++ b/examples/detect_simple.py
@@ -1,0 +1,241 @@
+"""
+    - do a straight coadd and store it in an ngmix Observation for the
+    simple_sim, which makes perfectly registered images with a pixel scale and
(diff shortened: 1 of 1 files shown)


esheldon:
It works

detect-example

esheldon:
I agree, we need to think how we will actually use all this.

Also, where will the analysis repo be to process the sims?  It doesn't have to be a "stack extension" at this stage where we are validating

beckermr:
yep all good questions!

PR #10, opened 2019-12-17
Add measurement code

esheldon:
This is still in progress

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ The doc strings for the main public APIs are complete. See them for more details
 
 ## SIM todo
 
-- [ ] create noise field
+- [x] create noise field
 - [ ] offset and rotate images
 - [ ] complex wcs in image and noise field
 - [ ] spatially varying PSF
@@ -45,7 +45,9 @@ The doc strings for the main public APIs are complete. See them for more details
 
 ## measure TODO
 
-- [ ] add tests of measure
+- [x] add noise field to CoaddObs
+- [ ] add test of CoaddObs, including noise and weight compatibility
+- [ ] add tests of SimMetadetect code
 - [ ] do real coadds for image and noise field
 - [ ] run centroiding algorithm in stack rather than using integer positions
 - [ ] run size measurement alg. in stack, to get postage stamp size, rather than using default footprint
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ The doc strings for the main public APIs are complete. See them for more details
 ## measure TODO
 
 - [x] add noise field to CoaddObs
-- [ ] add test of CoaddObs, including noise and weight compatibility
+- [x] add test of CoaddObs, including noise and weight compatibility
 - [ ] add tests of SimMetadetect code
 - [ ] do real coadds for image and noise field
 - [ ] run centroiding algorithm in stack rather than using integer positions
--- a/README.md
+++ b/README.md
@@ -45,6 +45,7 @@ The doc strings for the main public APIs are complete. See them for more details
 
 ## measure TODO
 
+- [ ] add tests of measure
 - [ ] do real coadds for image and noise field
 - [ ] run centroiding algorithm in stack rather than using integer positions
 - [ ] run size measurement alg. in stack, to get postage stamp size, rather than using default footprint
--- a/README.md
+++ b/README.md
@@ -47,10 +47,11 @@ The doc strings for the main public APIs are complete. See them for more details
 
 - [x] add noise field to CoaddObs
 - [x] add test of CoaddObs, including noise and weight compatibility
-- [ ] add tests of SimMetadetect code
-- [ ] do real coadds for image and noise field
+- [x] add smoke tests of SimMetadetect code
+- [ ] add substantial tests of SimMetadetect code
 - [ ] run centroiding algorithm in stack rather than using integer positions
 - [ ] run size measurement alg. in stack, to get postage stamp size, rather than using default footprint
+- [ ] do real coadds for image and noise field
 - [ ] evaluate stamp size is better, alg. based or footprint based
 - [ ] test with noise field propagated (need to add that in simple sim)
 - [ ] test with deblending and noise replacers
--- a/README.md
+++ b/README.md
@@ -34,3 +34,23 @@ for band_ind, band in enumerate(data):
 ## Documentation
 
 The doc strings for the main public APIs are complete. See them for more details.
+
+## SIM todo
+
+- [ ] create noise field
+- [ ] offset and rotate images
+- [ ] complex wcs in image and noise field
+- [ ] spatially varying PSF
+- [ ] defects with correct bit mask values, for image and noise field
+
+## measure TODO
+
+- [ ] do real coadds for image and noise field
+- [ ] run centroiding algorithm in stack rather than using integer positions
+- [ ] run size measurement alg. in stack, to get postage stamp size, rather than using default footprint
+- [ ] evaluate stamp size is better, alg. based or footprint based
+- [ ] test with noise field propagated (need to add that in simple sim)
+- [ ] test with deblending and noise replacers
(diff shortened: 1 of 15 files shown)


descwl_shear_testing/coadd_obs.py
```diff
@@ -0,0 +1,99 @@
...
+        image *= 1.0/wsum
+
+        self._psf = ngmix.Observation(
+            image=psf_image,
+            weight=psf_weight,
+            jacobian=jac,
+        )
+        self._image = image
+        self._weight = weight
+        self._jacobian = jac
```
beckermr: We would want all of these properties to be galsim Images in order to be consistent with the other APIs in the repo. We could alternatively switch those APIs to use numpy arrays.

descwl_shear_testing/coadd_obs.py
```diff
@@ -0,0 +1,99 @@
...
+        return self._jacobian
+
+    @property
+    def psf(self):
+        """
+        getter for psf observation
+        """
+        return self._psf
+
+    def _make_coadd(self):
```
beckermr: I think maybe this should be a function that returns another SEObs (or more general renamed class) with the coadded data.

esheldon:
@beckermr note I'm gonna rewrite almost all of this probably so ignore

descwl_shear_testing/mbobs_extractor.py
```diff
@@ -0,0 +1,437 @@
...
+        List of psf images
+    """
+
+    def __init__(self, *, config, exposures, sources, psfs=None):
+        self.config = config
+        self.exposures = exposures
+        self.sources = sources
+        self.psfs = psfs
+
+        self.log = lsst.log.Log.getLogger("MBObsExtractor")
```
beckermr: I am trying to learn to speak stack. What is going on here?

esheldon:
I ran a test with the current code, which is falling back to the integer location of objects from the peak finder.  This crude location will make the measurements noisy but not bias them, in principle.

```
m: -0.005 +/- -0.006
```

descwl_shear_testing/mbobs_extractor.py
```diff
@@ -0,0 +1,437 @@
...
+    def _get_mbobs(self, rec, weight_type='weight'):
+        """
+        make an ngmix.MultiBandObsList for input to the fitter
+
+        parameters
+        ----------
+        images: dict
+            A dictionary of image objects
+        rec: object record
+            TODO I don't actually know what class this is
```
beckermr: :)

descwl_shear_testing/mbobs_extractor.py
esheldon: this is setting the logger to show that as a prefix for its messages

descwl_shear_testing/mbobs_extractor.py
```diff
@@ -0,0 +1,437 @@
...
+        """
+
+        assert weight_type == 'weight'
+
+        # rec = self.sources[iobj]
+
+        mbobs = ngmix.MultiBandObsList()
+
+        for im_index, imf in enumerate(self.exposures):
+            # TODO: run noise replacers here
```
beckermr: If we do that here, will we need to do it on the noise images for cancelling correlated noise?

descwl_shear_testing/mbobs_extractor.py
esheldon: yeah, there are a few of these that I don't know yet

esheldon:
whoah, the response is negative:   -0.42

descwl_shear_testing/mbobs_extractor.py
esheldon: Yes, but I'm not sure exactly how this will affect us.

The deblended results, with noise replacement, look pretty bizarre much of the time

descwl_shear_testing/mbobs_extractor.py
```diff
@@ -0,0 +1,422 @@
...
+        row = cen.getY()
+        col = cen.getX()
+
+        wcs = imobj.getWcs().linearizePixelToSky(
+            orig_cen,
+            geom.arcseconds,
+        )
+        jmatrix = wcs.getLinear().getMatrix()
+
+        jacob = ngmix.Jacobian(
```
esheldon: I must have the convention wrong here and this is causing the negative responses

esheldon:
what is odd is that this is the same convention I used for the metacal extension we are using in lsst and it shows positive R

esheldon:
More stats
```
m = 0.0004 +/- 0.0030
```

esheldon:
by the way, these tests are without propagating a noise image, which is fine for this sim with no real coadding, interpolation, etc.  also it is very low noise anyway

esheldon:
more.
```
m: -7.05e-05 +/- 0.00077
```
probably enough for now

esheldon:
working fine with the noise image propagating to the shear measurement, this time with more noise
```
m: -0.002 +/- 0.002
```

esheldon:
I think this might be enough work on this particular PR

We will update the measurement code when we get something from bob to do centroiding with the stack.

In the mean time we can start other PRs for things like shifting and rotating images and doing proper coadds.

@beckermr do you want to review this?

beckermr:
Yeah I’ll take a look. I can say from having watched it being written that I’d prefer more tests.

esheldon:
argh, yes, sorry.  I forgot I had actually planned to add a bunch of tests.  I'll work on it.

README.md
```diff
@@ -38,7 +38,7 @@ The doc strings for the main public APIs are complete. See them for more details
 ## SIM todo
 
 - [x] create noise field
-- [ ] offset and rotate images
+- [ ] offset and rotate images, using Bob's code as an example
```
beckermr: where is this code?

esheldon: you pointed me to it on slack

beckermr: ohhhh that one

esheldon:
We should pick on the changes to the sim code and move the rest to a different repo

descwl_shear_testing/simple_sim.py
```diff
@@ -205,26 +206,32 @@ def gen_sim(self):
...
+                se_image += self._generate_noise_image(band_ind)
 
-                se_image += (
-                    self._noise_rng.normal(size=(self.se_dim, self.se_dim)) *
-                    self.noise_per_band[band_ind])
+                # make galsim image with same wcs as se_image but
+                # with pure random noise
+                noise_example = self._generate_noise_image(band_ind)
+                noise_image = se_image.copy()
+                noise_image.array[:, :] = noise_example
```
beckermr: Is this the preferred way to do this? I assume yes, but I don't actually know.

descwl_shear_testing/tests/test_simple_sim.py
```diff
@@ -11,7 +11,23 @@ def test_simple_sim_smoke():
...
+    data = sim.gen_sim()
+    for band in sim.bands:
+        for epoch in range(sim.epochs_per_band):
+            epoch_obs = data[band][epoch]
+
+            assert epoch_obs.noise is not None
+
+            nstd = epoch_obs.noise.array.var()
+            expected_std = 1/epoch_obs.weight.array[0, 0]
+            assert abs(nstd/expected_std-1) < 0.01
```
beckermr: These are variances not stds. IDK if you care, but the variable names are off.

setup.cfg
```diff
@@ -1,2 +1,3 @@
 [flake8]
 max-line-length = 88
+ignore =  N806,E226,W504
```
beckermr: I'd prefer not to have these warnings commented out globally. I know you want to keep the capital `T`, but my guess is that having flake8 check more things is better than being able to use `T`. However, I am not going to push you here, just stating my point of view.

descwl_shear_testing/coadd_obs.py
esheldon: I used this because the metadetect code expects an observation

PR #16, opened 2020-01-07
Tanwcs

esheldon:


esheldon:
this is currently erroring because in the render `toImage` requires a `CelestialCoord` for the `TanWCS`

beckermr:
Yeah I was going to get to this later this week. We should not be inheriting from the sims we have because of the API issues. Instead we should rewrite the base sim to always use celestial WCS conventions.

beckermr:
Supporting both flat and celestial stuff is just too much of a pain and puts too many branches in weird places.

esheldon:
Note we need to generate object ra, dec uniformly on the sky rather than uniform in x, y

I added the `util.randsphere` code to do this

PR #17, opened 2020-01-08
ENH TanWCS in sims

### LINKED ISSUE #12: build sims with a variable WCS
see the PR title

### LINKED ISSUE #14: use TanWCs in sims

### LINKED ISSUE #15: toImage should be given a CelestialCoord
In `render_objs_with_psf_shear` the position being sent should be a `CelestialCoord` not a `PositionD`

beckermr:
We use TanWCS everywhere now.

Closes #12
Closes #14 
Closes #15 
Closes #16


esheldon:
Does uniform in u, v equal uniform in ra, dec? are there cos(dec) terms needed?

beckermr:
UV is in the tangent plane so I think it is close but not exact.

beckermr:
See the math here: https://github.com/LSSTDESC/Coord/blob/master/coord/celestial.py#L637

beckermr:
tan is gnomic

beckermr:
Ack no I think I am wrong.

beckermr:
Alright, let me try to answer this question again without seeming like a crazy person. The projection that creates `(u, v)` from `(ra, dec)` is here: http://mathworld.wolfram.com/GnomonicProjection.html. In the code we are using, the math is here: https://github.com/LSSTDESC/Coord/blob/master/coord/celestial.py#L638

To make things simpler, let's suppose we are working with the projection centered at `ra=0, dec=0`. Then I get

```
u = cos(dec) * sin(ra) / c = tan(ra)
v = sin(dec) / c = tan(dec) / cos(ra)
c = cos(dec) * cos(ra)
```

So yeah, uniform in (u, v) locally does not appear at first glance to have the right derivatives to produce uniform points on the sphere.

beckermr:
OK @esheldon. Here is what we are doing now. For each TanWCS, there is a point on the image where it is formally tangent to the sphere. This in celestial coords is `wcs.center`. 

So to generate the dither for the random WCS, I 

1. generate a uniform dither in a coordinate system called `(u,v)` (units of arcsec, in the tangent plane centered at `wcs.center`). 
2. use `wcs.jacobian(world_pos=wcs.center).getMatrix()` to get the jacobian matrix relating `(u,v)` to `(x,y)` where `(x,y)` is in the image. This jacobian is the same as those used in ngmix etc.
3. invert this jacobian matrix and then turn the dither in `(u,v)` into a dither in `(x,y)`
4. rebuild the `TanWCS` with the origin shifted by this dither

For shearing the scene, I 

1. compute the `(u, v)` position of the object (TBD, but likely take ra-dec pos, got to xy, compute dx, dy from image pos of `wcs.center` and then use jacobian at `wcs.center` to convert dx, dy to du, dv)
2. shear it using the jacobian matrix corresponding to the shear (`shear.getMatrix()`)
3. then undo 1 back to (ra, dec)

For rendering, I use the jacobian at the final object position on the sphere.

cc @rmjarvis for viz and comments

beckermr:
Alright @esheldon @rmjarvis @rearmstr, sims with a variable WCS appear to be working! 

download-1

I am still not 100% sure exactly what we want to mean by constant shear on the sphere. I can think of two meanings.

1. A sheared round object will have the same angle with respect to the local ra-dec lines at the object's center. 

2. If you parallel transported a sheared round object to the center of the patch along a geodesic, they would all have the same shape. 

These two things are not the same IIUIC. I have currently implemented the first one, since using codes we have now, this will result in a catalog of shapes that should be all the same (i.e. our current codes report shears wrt the local ra-dec basis.) 

However, from a physical point of view, the second is absolutely more correct (and would better match assumptions about shearing the scene in metacal). The issue with it is that if we use it, then our codes will not report a constant shear and we will have to remove the effect of the coordinate system before reporting shear measurement biases. This is not hard (you parallel transport your shapes from their measured locations back to the center of the patch), but I am punting on this now.

Any comments? Also some code review would be great!

esheldon:
why do some of the images show objects highly stretched along the rotated axis?

beckermr:
I included a shear in the wcs

esheldon:
I see; how large were the variations in shear?  Can I have a look at the code that generated these?

esheldon:
I found the notebook; shear std 0.1

esheldon:
I ran a quick test with a single epoch, just to make sure we can still recover the shear, and it looks ok
```
m: -0.004 +/- 0.003
```

PR #20, opened 2020-01-15
ENH add wldeblend gals

beckermr:
This PR adds more realistic gals to the sim. cc @esheldon for viz


descwl_shear_sims/simple_sim.py
```diff
@@ -284,7 +392,10 @@ def gen_sim(self):
                 noise_image = se_image.copy()
                 noise_image.array[:, :] = noise_example
 
-                se_weight = se_image.copy() * 0 + 1.0 / self.noise_per_band[band_ind]**2
+                se_weight = (
+                    se_image.copy() * 0
+                    + 1.0 / self.noise_per_epoch[band_ind]**2
```
beckermr: @esheldon This was a bug in the sim noise fields. Fixed now.

esheldon: ok.  But should not have caused any problems, because the noise actually added was the same as that used to generate the noise fields for metacal

beckermr: yah for sure - we just want to get the depth of the sim right.

PR #21, opened 2020-01-16
ENH bigger default buffers

beckermr:
This PR fixes the detection issues with stack.


esheldon:
probably need slightly looser tolerance on that check

Issue #27, opened 2020-01-21
turning on cosmics we are getting zero detections in 15% of coadds


esheldon:

esheldon:
I meant zero detections in the coadds

beckermr:
Do you have example images?

esheldon:
This reproduces it on the second trial
```bash
python ~/git/metadetect/examples/lsst_sim.py \
    --cosmic-rays \
    --nepochs 1 \
    --noise 1 \
    --ntrial 2 \
    --seed 244714 \
    --output /tmp/test.fits
```

esheldon:
Here you can see that a hot 10^18 pixel is left at the bottom
![example](https://user-images.githubusercontent.com/621661/72803999-70a80880-3c1d-11ea-8563-19ef60671661.png)

esheldon:
that is, after "repair"

beckermr:
Ahhhhh. So the cosmic ray finder misses the pixel and then the detections get mucked up.

We can turn down the cosmic ray level to "saturation" but this very well could be an issue in real data. 

We probably need an outlier masking step like we use in DES.

esheldon:
the resulting coadd ends up looking like flat white image

beckermr:
We should consult the stack folks on this. They must have processing steps that do this sort of thing.

esheldon:
Its probably because it is near the edge; I bet they typically mask those areas

esheldon:
(pure speculation)

beckermr:
They have edge bit mask flags in the code. Maybe we should use them...

esheldon:
before repair

![before](https://user-images.githubusercontent.com/621661/72804308-1e1b1c00-3c1e-11ea-93c0-fd24dad0183b.png)

after repair

![after](https://user-images.githubusercontent.com/621661/72804332-27a48400-3c1e-11ea-86d0-133d9245a2bf.png)

esheldon:
there is a derivative involved with the CR finding, probably cannot be done at the edge

beckermr:
We could also only draw cosmics in the center

beckermr:
https://github.com/LSSTDESC/descwl-shear-sims/issues/28

esheldon:
Yeah, that would probably fix it.  Or I can zero out edges in the coadd code before repair, which is a good test in any case.  I'll try it.

esheldon:
[...]

PR #29, opened 2020-01-21
Turn on edge bits

esheldon:
set the EDGE bit for a border around the image

### DIFF
--- a/descwl_shear_sims/simple_sim.py
+++ b/descwl_shear_sims/simple_sim.py
@@ -283,7 +283,9 @@ class Sim(object):
 
         # the SE image could be rotated, so we make it big enough to cover the
         # whole coadd region plus we make sure it is odd
-        self.se_dim = int(np.ceil(self.coadd_dim * np.sqrt(2))) + 10
+        self.se_dim = (
+            int(np.ceil(self.coadd_dim * np.sqrt(2))) + 10 + 2*edge_width
+        )
         if self.se_dim % 2 == 0:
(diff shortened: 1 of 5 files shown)


esheldon:
Note it turns out this does not address #27

esheldon:
I added a test; what do you have in mind?

descwl_shear_sims/simple_sim.py
```diff
@@ -27,10 +34,6 @@
     'wldeblend': {'ngals_factor': 1.0},
 }
 
-# default mask bits from the stack
-BAD_COLUMN = np.int64(2**0)
-COSMIC_RAY = np.int64(2**3)
-
 
 # double check they match the stack
 def _check_bits_against_stack():
```
beckermr: This function should go in the lsst_bits.py module.

beckermr:
This test needs updating: https://github.com/LSSTDESC/descwl-shear-sims/blob/master/descwl_shear_sims/tests/test_lsst_mask_bits.py

descwl_shear_sims/simple_sim.py
```diff
@@ -90,6 +68,8 @@ class Sim(object):
     buff : int, optional
         The width of the buffer region in the coadd image where no objects are
         drawn. Default is 50.
+    edge_width:  int
+        Width of boundary to be marked as EDGE in the bitmask
```
beckermr: I'd be nice to have the default in the doc string.

PR #35, opened 2020-01-27
add support for doing our own psf coadd

esheldon:


esheldon:
I want to leave it because it is very possible that we will use it later if we figure out how to give the stack a psf that can do sub-pixel rendering and it does in fact recenter the psf

descwl_shear_sims/simple_sim.py
```diff
@@ -298,8 +300,28 @@ def __init__(
         LOGGER.info('area is %f arcmin**2', self.area_sqr_arcmin)
 
         # info about coadd PSF image
-        self.psf_dim = 53
+        self.coadd_psf_dim = 31
```
beckermr: This is pretty small, right? Pixels here are 0.2 arcsec.

descwl_shear_sims/simple_sim.py
```diff
@@ -657,24 +679,38 @@ def _get_gal_wldeblend(self):
...
+            position_angle_range=self.wcs_kws.get('position_angle_range', (0, 0)),
+            dither_range=self.wcs_kws.get('dither_range', (0, 0)),
+            scale_frac_std=self.wcs_kws.get('scale_frac_std', 0),
+            shear_std=self.wcs_kws.get('shear_std', 0),
+            scale=self.scale,
+            world_origin=self._world_origin,
+            origin=self._se_origin,
+            rng=self._rng,
+        )
+        wcs_kws['psf_origin'] = self._psf_origin
```
beckermr: Why is this not in the full `dict` creation statement above?

descwl_shear_sims/simple_sim.py
```diff
@@ -706,10 +743,13 @@ def _psf_render_func(*, x, y, center_psf):
...
                     y=y-int(y+0.5),
                 )
-            return psf.drawImage(
+            psf_image = psf.drawImage(
                 nx=self.psf_dim,
                 ny=self.psf_dim,
                 offset=offset,
-                wcs=se_wcs.local(image_pos=image_pos))
+                wcs=psf_wcs,
+                # wcs=se_wcs.local(image_pos=image_pos),
```
beckermr: Why don't we need the `.local` call to make sure we get the right WCS at the PSF location?

descwl_shear_sims/tests/test_gen_tanwcs.py
```diff
@@ -27,15 +30,18 @@ def test_gen_tanwcs_dither():
         ra=20 * galsim.degrees,
         dec=-10 * galsim.degrees)
     origin = galsim.PositionD(x=10, y=11)
-    wcs = gen_tanwcs(
+    psf_origin = galsim.PositionD(x=10, y=11)
```
beckermr: Let's make this center different so that we can test that the two origins are not swapped.

descwl_shear_sims/tests/test_gen_tanwcs.py
```diff
@@ -52,22 +58,28 @@ def test_gen_tanwcs_dither():
     assert np.allclose(wcs.crpix[0], origin.x + dxdy[0])
     assert np.allclose(wcs.crpix[1], origin.y + dxdy[1])
 
+    assert np.allclose(psf_wcs.crpix[0], origin.x + dxdy[0])
+    assert np.allclose(psf_wcs.crpix[1], origin.y + dxdy[1])
```
beckermr: This test is wrong. It should be psf_origin.

descwl_shear_sims/tests/test_gen_tanwcs.py
```diff
@@ -52,22 +58,28 @@ def test_gen_tanwcs_dither():
...
+
 
 def test_gen_tanwcs_jacobian():
     seed = 42
     world_origin = galsim.CelestialCoord(
         ra=20 * galsim.degrees,
         dec=-10 * galsim.degrees)
     origin = galsim.PositionD(x=10, y=11)
-    wcs = gen_tanwcs(
+    psf_origin = galsim.PositionD(x=10, y=11)
```
beckermr: Ditto here for making the PSF origin different in the test.

descwl_shear_sims/tests/test_gen_tanwcs.py
```diff
@@ -98,9 +112,11 @@ def test_gen_tanwcs_seed():
         world_origin=galsim.CelestialCoord(
             ra=20 * galsim.degrees,
             dec=-10 * galsim.degrees),
-        origin=galsim.PositionD(x=10, y=11))
+        origin=galsim.PositionD(x=10, y=11),
+        psf_origin=galsim.PositionD(x=10, y=11),
```
beckermr: Again make the origin different.

descwl_shear_sims/simple_sim.py
esheldon: Does a local wcs know anything about the origin? That info is needed to do the coadd

beckermr: Well galsim when drawing an image calls .local anyways. 

All of this code actually needs to be in the coaddin package. There you should create the coadd wcs you need, draw the psf models at the appropriate location using the sims in modified, and then coadd them.

beckermr:
We won’t always have a constant PSF, so we should not merge this. We can attach a psf coadd wcs for use later, but we need to keep the rest of the changes out.

esheldon:
I know its all wrong, but I needed a way to develop on a local machine and then deploy on a remote machine.

This required making a branch

Since I had a branch, I decided to do a PR just in case we wanted to pursue this general line, and to open up to comments.  I would not merge until we had something that worked in general.

PR #38, opened 2020-01-30
Psf offsets

### LINKED ISSUE #37: optionally return offset for psf image
This is needed for the psf coaddition

esheldon:
closes #37

### DIFF
--- a/descwl_shear_sims/simple_sim.py
+++ b/descwl_shear_sims/simple_sim.py
@@ -696,7 +696,7 @@ class Sim(object):
             else:
                 raise ValueError('psf_type "%s" not valid!' % self.psf_type)
 
-        def _psf_render_func(*, x, y, center_psf):
+        def _psf_render_func(*, x, y, center_psf, get_offset=False):
             image_pos = galsim.PositionD(x=x, y=y)
             psf = _psf_galsim_func(x=x, y=y)
             if center_psf:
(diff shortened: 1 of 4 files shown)


beckermr:
See my comment. This PR is not needed because the user already has the info they need for this. The offset is simply `x - int(x+0.5), y - int(y+0.5)`.

descwl_shear_sims/simple_sim.py
```diff
@@ -706,10 +706,15 @@ def _psf_render_func(*, x, y, center_psf):
                     x=x-int(x+0.5),
                     y=y-int(y+0.5),
```
beckermr: The offset is here and directly computed from the inputs. No need to send it back out.

esheldon: I want this available so I don't have to hard code `x=x-int(x+0.5)` in other codes that use the offset.  What if conventions change and the offset is calculated differently in the future? etc.

esheldon: Also, it does no harm to have it there.

beckermr: This is the convention for a pixel center. It won't change and hasn't in probably 20+ years. :)

beckermr: Feel free to merge if you want though.

esheldon: A real case where the convention could be different:  what if the stack uses pixel centers that are 0.5 different from galsim?

PR #39, opened 2020-01-31
Psf shape

esheldon:
Add ability to send the psf shape in the psf_kws

Add a test

### DIFF
--- a/descwl_shear_sims/simple_sim.py
+++ b/descwl_shear_sims/simple_sim.py
@@ -252,8 +252,8 @@ class Sim(object):
         self.psf_kws = copy.deepcopy(PSF_KWS_DEFAULTS[self.psf_type])
         if psf_kws is not None:
             self.psf_kws.update(psf_kws)
-        self.psf_g1 = self.psf_kws.pop('psf_g1', 0.0)
-        self.psf_g2 = self.psf_kws.pop('psf_g2', 0.0)
+        self.psf_g1 = self.psf_kws.pop('g1', 0.0)
+        self.psf_g2 = self.psf_kws.pop('g2', 0.0)
 
         self.wcs_kws = wcs_kws or {}
 
--- a/descwl_shear_sims/simple_sim.py
+++ b/descwl_shear_sims/simple_sim.py
@@ -32,6 +32,9 @@ GAL_KWS_DEFAULTS = {
     'exp': {'half_light_radius': 0.5},
     'wldeblend': {'ngals_factor': 1.0},
 }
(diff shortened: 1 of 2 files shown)


descwl_shear_sims/simple_sim.py
```diff
@@ -246,7 +249,12 @@ def __init__(
         self._final_gal_kws = gal_kws_defaults
 
         self.psf_type = psf_type
-        self.psf_kws = psf_kws or {'fwhm': 0.8}
+        self.psf_kws = copy.deepcopy(PSF_KWS_DEFAULTS[self.psf_type])
+        if psf_kws is not None:
+            self.psf_kws.update(psf_kws)
+        self.psf_g1 = self.psf_kws.pop('psf_g1', 0.0)
+        self.psf_g2 = self.psf_kws.pop('psf_g2', 0.0)
```
beckermr: Why does it say `psf_g1` if it is in the psf_kws in the first place? We can use `g1` only I think.

esheldon: good point

descwl_shear_sims/simple_sim.py
```diff
@@ -252,8 +252,8 @@ def __init__(
         self.psf_kws = copy.deepcopy(PSF_KWS_DEFAULTS[self.psf_type])
         if psf_kws is not None:
             self.psf_kws.update(psf_kws)
-        self.psf_g1 = self.psf_kws.pop('psf_g1', 0.0)
-        self.psf_g2 = self.psf_kws.pop('psf_g2', 0.0)
+        self.psf_g1 = self.psf_kws.pop('g1', 0.0)
+        self.psf_g2 = self.psf_kws.pop('g2', 0.0)
```
beckermr: My final comment is that we should probably just store the whole dictionary and then pull out what we need in the method, but I am willing to pass on that.

esheldon: The issue there is that we need to pop these because they are not allowed keywords to the Gaussian itself.

esheldon: So we would need to do this in the method to avoid popping the main dictionary
```
kw = deepcopy(self.psf_kws)
g1 = kw.pop('g1', 0.0)
g2 = kw.pop('g2', 0.0)
```
which would be fine too

beckermr: Right. Eventually, we can use the galsim parsing internals to build this automatically with any allowed keywords in a galsim config. However, let's merge for now and punt on that.

PR #53, opened 2020-02-04
add se_dim keyword

### LINKED ISSUE #52: add control over single epoch dims
with keywords, currently it is calculated from coadd dims

esheldon:
closes #52 

se_dim will be set, only check is that the image size isn't
smaller than 2*edge width

added test

### DIFF
--- a/descwl_shear_sims/simple_sim.py
+++ b/descwl_shear_sims/simple_sim.py
@@ -73,6 +73,9 @@ class Sim(object):
         drawn. Default is 50.
     edge_width:  int
         Width of boundary to be marked as EDGE in the bitmask, default 5
+    se_dim: int
+        Dimensions of the single epoch image.  If None (the default) the
+        size is chosen to encompass the coadd for small dithers.
     ngals : float, optional
         The number of objects to simulate per arcminute^2. Default is 80.
     grid_gals : bool, optional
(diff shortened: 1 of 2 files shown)


esheldon:
ok with you @beckermr ?

descwl_shear_sims/simple_sim.py
```diff
@@ -262,13 +266,18 @@ def __init__(
...
-        # the SE image could be rotated, so we make it big enough to cover the
-        # whole coadd region plus we make sure it is odd
-        self.se_dim = (
-            int(np.ceil(self.coadd_dim * np.sqrt(2))) + 10 + 2*edge_width
-        )
-        if self.se_dim % 2 == 0:
-            self.se_dim = self.se_dim + 1
+        if se_dim is not None:
+            self.se_dim = int(se_dim)
+            assert self.se_dim > 2*edge_width
```
beckermr: Shouldn't this lower limit be bigger?

esheldon: I guess we could add an arbitrary number to it

beckermr: Well we have the coadd dim and we might have rotations, so I think the default is the min.

esheldon: But the point of the se_dim keyword is so the user can set it to whatever size they want, even something that could have an edge in it and get rejected later

beckermr: But this keyword is set for all epochs. Why would we want to reject all of them?

esheldon: good point

PR #57, opened 2020-02-18
Star masks

### LINKED ISSUE #43: bright star masks
For now we could perhaps adopt ones from DES, but they will I think be generally different for LSST.

For one thing the are obliterated by a star will be larger for LSST

### LINKED ISSUE #44: bleed trail masks
Can maybe use DES ones for now, but they will differ for LSST

esheldon:
closes #43 
closes #44 

At this point there are heuristics for star radii and bleed trail properties, but we can alter it to take in those values if we get a catalog of these from DC2

I defined bits for these, there are no bits in the official lsst code

TODO is to add this to the simple_sim code with keywords star bool and star_kws

### DIFF
--- a/descwl_shear_sims/gen_star_masks.py
+++ b/descwl_shear_sims/gen_star_masks.py
@@ -1,23 +1,13 @@
 import numpy as np
 from numba import njit
-import esutil as eu
+import galsim
 import ngmix
+from .randsphere import randcap
 
 
-def add_star_and_bleed_masks(*,
-                             rng,
-                             mask,
-                             wcs,
-                             density=200,
-                             pixel_scale=0.2,
-                             radmean=3,
-                             radstd=5,
-                             radmin=3,
-                             radmax=50,
-                             bleed_length_fac=2,
-                             mask_value=1):
+class StarMasks(object):
     """
-    make fake masks with
+    make fake masks around the given position with
 
     - star mask radii distributed as a clipped log normal
     - bleed trail length a fixed factor larger than the diameter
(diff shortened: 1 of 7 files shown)


descwl_shear_sims/lsst_bits.py
```diff
@@ -6,6 +6,11 @@
 COSMIC_RAY = np.int64(2**3)
 EDGE = np.int64(2**4)
 
+# these are not official values, as far as I know
+# the stack doesn't mark such things (ES 2020-02-18)
+STAR = np.int64(2**10)
+BLEED = np.int64(2**11)
```
beckermr: Do we want to go higher here so things don't break later?

descwl_shear_sims/tests/test_gen_star_masks.py
```diff
@@ -0,0 +1,82 @@
+"""
+copy-paste from my (beckermr) personal code here
+https://github.com/beckermr/metadetect-coadding-sims
+"""
```
beckermr: This comment is wrong.

descwl_shear_sims/tests/test_gen_star_masks.py
```diff
@@ -0,0 +1,82 @@
...
+import numpy as np
+
+from ..gen_star_masks import StarMasks
+from ..lsst_bits import STAR, BLEED
+from ..simple_sim import Sim
+
+
```
beckermr: We should add a test here that the code is repeatable if we call it twice with the same rng seed.

descwl_shear_sims/lsst_bits.py
esheldon: We will definitely be changing these so it doesn't matter too much.

It looks like the stack masks are 32 bit so we could use 29 and 30

descwl_shear_sims/tests/test_gen_star_masks.py
esheldon: added

descwl_shear_sims/tests/test_gen_star_masks.py
esheldon: removed

descwl_shear_sims/gen_star_masks.py
```diff
@@ -26,24 +26,25 @@ class StarMasks(object):
     center_dec: float
         Stars will be generated in a disk around this point
     radius_degrees: float
-        Stars will be in a disk with this radius
+        Stars will be generated in a disk with this radius around the coadd
+        center, default 5 degrees
     density: float
-        Number of saturated stars per square degree
-    pixel_scale: float
-        pixel scale in arcsec/pixel
```
beckermr: Are we removing this option? I don't see the docs added back.

Issue #60, opened 2020-02-20
merge stars and saturated stars


esheldon:
the mechanism for saturated stars is different from galaxies and normal stars

So currently there is no star drawn where there is saturation added.

Figure out how to merge these

PR #61, opened 2020-02-20
Stars

### LINKED ISSUE #59: add stars
this is in addition to the saturated stars which are separate, but we may want to merge the lists of positions and draw stars before setting the values to SAT_VAL

esheldon:
closes #59


beckermr:
Shouldn't the star masks know about where the stars are?

esheldon:
See #60

esheldon:
I'm going to merge this and do a separate PR to combine the sat stars and regular stars

PR #66, opened 2020-02-25
combine sat stars with regular stars

### LINKED ISSUE #60: merge stars and saturated stars

esheldon:
Whether or not the star is saturated is stored in the dict
for the object

closes #60

### DIFF
--- a/descwl_shear_sims/render_sim.py
+++ b/descwl_shear_sims/render_sim.py
@@ -7,7 +7,6 @@ import logging
 
 import numpy as np
 import galsim
-from .gen_star_masks import add_star_and_bleed
 
 LOGGER = logging.getLogger(__name__)
 
@@ -59,16 +58,11 @@ def render_objs_with_psf_shear(
         xmin=0,
         ymin=0,
     )
-    mask_im = galsim.ImageI(
-        nrow=img_dim,
-        ncol=img_dim,
-        xmin=0,
-        ymin=0,
-    )
 
     jac_wcs = wcs.jacobian(world_pos=wcs.center)
     center_image_pos = wcs.toImage(wcs.center)
 
+    overlap_info = []
     for obj_data, uv_offset, in zip(objs, uv_offsets):
 
         obj = obj_data['obj']
@@ -124,19 +118,14 @@ def render_objs_with_psf_shear(
         overlap = stamp.bounds & se_im.bounds
         oshape = overlap.numpyShape()
         if oshape[0]*oshape[1] > 0:
+            overlaps = True
             se_im[overlap] += stamp[overlap]
+        else:
+            overlaps = False
 
-            if (obj_data['type'] == 'star' and obj_data['saturated']):
+        overlap_info.append({
+            'overlaps': overlaps,
+            'pos': pos,
+        })
 
-                sat_data = obj_data['sat_data']
-                add_star_and_bleed(
-                    mask=mask_im.array,
-                    image=se_im.array,
-                    x=pos.x,
-                    y=pos.y,
-                    radius=sat_data['radius'],
-                    bleed_width=sat_data['bleed_width'],
-                    bleed_length=sat_data['bleed_length'],
-                )
-
-    return se_im, mask_im
+    return se_im, overlap_info
--- a/descwl_shear_sims/render_sim.py
+++ b/descwl_shear_sims/render_sim.py
@@ -7,6 +7,7 @@ import logging
 
 import numpy as np
 import galsim
+from .gen_star_masks import add_star_and_bleed
 
 LOGGER = logging.getLogger(__name__)
 
@@ -49,10 +50,21 @@ def render_objs_with_psf_shear(
     se_image : galsim.ImageD
         The rendered image.
     """
+
     shear_mat = galsim.Shear(g1=g1, g2=g2).getMatrix()
 
     se_im = galsim.ImageD(
-        nrow=img_dim, ncol=img_dim, xmin=0, ymin=0)
+        nrow=img_dim,
+        ncol=img_dim,
+        xmin=0,
+        ymin=0,
+    )
+    mask_im = galsim.ImageI(
+        nrow=img_dim,
+        ncol=img_dim,
+        xmin=0,
+        ymin=0,
+    )
 
     jac_wcs = wcs.jacobian(world_pos=wcs.center)
     center_image_pos = wcs.toImage(wcs.center)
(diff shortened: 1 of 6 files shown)


descwl_shear_sims/render_sim.py
```diff
@@ -114,4 +126,17 @@ def render_objs_with_psf_shear(
         if oshape[0]*oshape[1] > 0:
             se_im[overlap] += stamp[overlap]
 
-    return se_im
+            if (obj_data['type'] == 'star' and obj_data['saturated']):
```
beckermr: I'm not a huge fan of having this done here in this function. The only dependence needed is the star position. It seems like we should be able to abstract this out.

esheldon: do you mean factor this out?  What do you suggest?

descwl_shear_sims/simple_sim.py
```diff
@@ -242,11 +244,8 @@ class Sim(object):
     sat_stars_kws : dict, optional
         A dictionary of options for generating star and bleed trail masks
 
-            radius_degrees: float
-                Stars will be generated in a disk with this radius around
-                the coadd center, default 5 degrees
             density: float
-                Number of saturated stars per square degree, default 200
+                Number of saturated stars per square arcmin, default 0.055
```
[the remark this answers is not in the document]
esheldon: That is just the density of saturated stars, much lower density than the 1/sq arcmin of ordinary stars (psf stars are also more rare  than 1/sq arcmin ordinary stars).

I just got a rough estimate of 200/sq degree from eyeballing some images, we can certainly tune this better

descwl_shear_sims/render_sim.py
esheldon: The bleeds should be part of rendering, since they are rendering

Do you no like setting a mask here?

beckermr: this function is for render galaxies etc. - it setting mask bits is weird to me and that will make it harder to test - we should have a separate function for that and we should test it properly

esheldon: The argument against that is it will involve duplicating a bunch of code that is within that loop.

To avoid that I would want to refactor all that stuff into a separate function

esheldon: As it is currently written, we would have to re-render the stamps, or save them for later, in order to do the overlap calculation

I guess one way around that is to render the bleed in the case that the center is located in the image rather than just when there is overlap, but that would mean that saturation would only occur in some of the epochs

esheldon: To put it more clearly:  that would mean that there are cases where there should be saturation but there isn't

esheldon: a better option would be to save a boolean whether there was overlap

beckermr: Im confused by these comments. The code doing the wcs transforms in the loop is general enough that we could abstract it. However I don’t want this to be a sticking point. My main concern is testing. I wrote a bunch of tests of this function and now with an extra branch we will need more of them.

esheldon: It also involves refactoring the entire code base for bright star masking, because I combine the adding of saturation and mask setting there

beckermr: Yeah that refactor should be done anyways to be honest. We are spraying and mangling state all over the place at some level.

esheldon: Look at this function:

https://github.com/LSSTDESC/descwl-shear-sims/blob/master/descwl_shear_sims/gen_star_masks.py#L224

Why would we write two separate functions, each of which do this same book keeping and calculations, with the only difference being in the last line?

esheldon: I'm going to take this stuff out of the render function because I think its fine, but I don't agree that we should duplicate code like the above

beckermr: Right. We should abstract the wcs handling too.

beckermr: If we are going to store a dict of info for each object, we can double down on that and store the wcs image positions and Jacobians there. Then each render function can consume that info.

esheldon: How about the recent commits?

beckermr: Right this is closer. My thought is to take the WCS code and stamp computation code here: https://github.com/LSSTDESC/descwl-shear-sims/pull/66/files#diff-5873b4ec62708400cfd519b1e39dfc5bR87 and move it to a separate function. 

Then we run this function once per object and epoch combination. The results we need get stored in the dict attached to each object. Then we feed that data back into a separate function that renders the object and another function the builds the masks. 

This separates out a lot of the code and keeps the code paths pretty linear and independent.

beckermr: I know I am pushing hard on the code structure here. My intent is that this code will live with us for a long time, so I want to get it right now.

esheldon: We still cannot do the overlap calculation until we render the stamp

beckermr: So we do a prerender of the stamp that is really fast to get the stamp size. This doesn't actually render the object. It is this part we should move out.

beckermr: From that and the extra info, we can compute the stamp bounds and overlap.

esheldon: We don't want to add them to the object dict which is per band not per epoch, unless we restructure that as well

beckermr: Well, we can store anything there, so as long as we are keeping extra state around, we might as well codify that into a formal API that we use consistently.

esheldon: I'm looking at this code and it just isn't obvious to me how to do what you are asking.  It is your code not mine so it isn't so easy to envision it.

I propose we merge this and open an new issue about refactoring, which you can work on when you get back from your leave. I doubt I will be implementing any new features that will require rendering beyond what we currently have:  at the most I plan to improve the drawing of saturated regions around stars, and improve bleed shapes, which fits in fine with the current setup.

esheldon: To be clear, I'm disagreeing with you, I just don't want to attempt it myself

esheldon: sorry, I'm *not* disagreeing :)

beckermr: Works for me!

Issue #76, opened 2020-03-02
the wldeblend images are on different zero points, but we are not passing this on


esheldon:
We need to pass this information to the coadding some how, or put them onto the same system

I expect that the real DM processing forces all images to have the same zero point

beckermr:
Wait what? How are the zero points different?

esheldon:
What's the mag zero point?

beckermr:
The issues title says the images are on different zero points. They should not be. I use whatever flux comes out of wldeblend so the zero point should always be the same, right?

esheldon:
Here is the formula used:
```python
self.exposure_time*self.zero_point*10**(-0.4*(ab_magnitude-24))
```
The leading terms are explicitly taking the image *off* of the zero point of 24 and putting it onto some kind of instrumental zero point

beckermr:
Yes but it is the same for **all** images, right?

Do you mean across different bands?

esheldon:
Yes, different bands are gettinga different zero point and these are not close to 24, they range widely

esheldon:
I plan to rescale them in the sim to zero point 30

PR #77, opened 2020-03-02
Realistic star mags

### LINKED ISSUE #73: draw stellar mags from the expected distribution
I have examples from Javier Sanchez

These are extrapolated from SDSS which seems highly suspect to me, but apparently these were approved for DC2

### LINKED ISSUE #74: deal with bounding boxes for big bright objects
They are getting clipped

clipped

esheldon:
closes #73 
closes #74 

It is a bit hard to check we are getting the right mags out for a number of reasons:

1. the zero points are all over the place due to wldeblend
2. we are not rescaling them all on the same system before coadding
3. we are coadding multiple bands

### DIFF
--- a/descwl_shear_sims/cache_tools.py
+++ b/descwl_shear_sims/cache_tools.py
@@ -1,0 +1,7 @@
+import functools
+import fitsio
+
+
+@functools.lru_cache(maxsize=8)
+def cached_catalog_read(fname):
+    return fitsio.read(fname)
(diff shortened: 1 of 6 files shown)


esheldon:
mag-hist

esheldon:
closes #76

esheldon:
This looks good to me, any comments @beckermr  ?

PR #83, opened 2020-03-04
Saturate images

### LINKED ISSUE #82: limit bright stars in some way
Including very bright stars can cause big biases

This can be dealt with by

1. Using saturation with a real clipped value and marking it saturated, in addition to whatever bleed effects we include
2. If saturation is not turned on, do not include the stars that would saturate

esheldon:
closes #82 

Also when not doing saturation, not including the bright stars, but may change this


esheldon:
TODO:

- [ ] get sat values for riz
- [ ] evaluate whether we should not include bright stars if sat_stars is False

PR #86, opened 2020-04-01
ENH refactor

beckermr:
cc @esheldon


beckermr:
OK @esheldon This one is ready for a good look. The code now supports grids for stars and galaxies. I have also normalized the API for both so that things are consistent. I added some tests and the rendering looks correct.

beckermr:
I've tagged master at 0.1 since the API is changing a bit and so code that calls this package will need changes.

descwl_shear_sims/simple_sim.py
```diff
@@ -923,84 +931,72 @@ def _generate_noise_image(self, band_ind):
...
+            all_data.append(obj_data)
 
-        return all_data, uv_offsets
+        return all_data
 
     def _keep_star(self, star):
+        """remove stars under certain conditions
 
+        the conditions are:
+          - too faint
```
esheldon, line 1002: too bright

descwl_shear_sims/simple_sim.py
```diff
@@ -101,13 +101,27 @@ class Sim(object):
...
+    layout_kws : dict, optional
+        A dictionary giving options for the specific layout type. These vary per
+        layout type and are
+
+            'random' - no options
+            'grid' - any of the following
+                'dim' : int, required
+                    An integer giving the dimension of the grid on one axis.
+
+    grid_objects : int, optional
```
esheldon, line 120: grid_objects sounds like a boolean but it is an integer

esheldon:
Can you please make available an example config file that has the defaults set, as well as commented parts for all the non-defaults that would work if uncommented?

descwl_shear_sims/simple_sim.py
beckermr: This one shouldn't even be here. Will remove.

descwl_shear_sims/simple_sim.py
beckermr:
```suggestion
          - too bright
```

beckermr:
OK @esheldon. See the file `config.yaml` for an example of the current options.

PR #88, opened 2020-04-07
make config use mainly defaults

esheldon:


esheldon:
I also added some additional docs

PR #91, opened 2020-04-09
generalize bright region marking

esheldon:
if any object has saturation it will get a bright mask
This is done for all bands even if sat is in one band

The motivation for this sort of thing is to give downstream codes enough information to avoid trying to deconvolve parts of the image that will cause artifacts.

If we had a reliable way to identify bright objects in the presence of saturation downstream we might be able to avoid putting this code in the sim, but I don't think we have that yet.  In general we might rely on external catalogs to do this in real data (all the bright objects are known).


esheldon:
This is working and I'm ready to merge.  There is still a TBD making these variable size rather than fixed size, but can address that later.

PR #93, opened 2020-04-14
BUG use correct noise for 10-year LSST depth

beckermr:
cc @esheldon


esheldon:
also need to correct examples/config.yaml where it says default is 100

PR #95, opened 2020-04-16
Galaxy builder

### LINKED ISSUE #94: implement our own GalaxyBuilder to make round galaxies
This will inherit from the descwl.GalaxyBuilder and allow forcing objects to be round

this should speed up tests with gals_type 'wldeblend' significantly

esheldon:
Add option to make wldeblend galaxies round

closes #94

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ data = Sim(rng=10, epochs_per_band=3).gen_sim()
 
 for band_ind, band in enumerate(data):
     for epoch_ind, obs in enumerate(data[band]):
-        # do something with obs here
+        # do something with obs here.
         # obs.image is the image
         # obs.weight is the weight map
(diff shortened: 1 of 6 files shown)


esheldon:
Its hard to add detailed tests for this one, since we don't have a way to control the object s/n or the background noise.  So we can't for example test if the objects are round

beckermr:
We can render one in the center of a sim and check that its moments matrix is symmetric.

esheldon:
We can't do that using a Sim; do you mean to do the calls to the galaxy builder directly?

beckermr:
Why? Just make a sim on a grid with one object.

esheldon:
can you do grid and wldeblend?

beckermr:
I think so yes.

esheldon:
hmm.... travis is not running

beckermr:
let me try something

PR #96, opened 2020-04-20
Check stars_kws and gals_kws for valid entries

esheldon:


descwl_shear_sims/simple_sim.py
```diff
@@ -576,12 +587,20 @@ def _setup_stars(
...
+                self.stars_kws,
+                ('density', 'mag', 'min_mag'),
+                'stars_kws',
+            )
+            if ('density' in self.stars_kws and
+                    isinstance(self.stars_kws['density'], dict)):
+                check_keys(
+                    self.stars_kws['density'],
+                    ('min_density', 'max_density'),
+                    'stars_kws["density"]',
```
beckermr, line 600: Can we pull the list of valid keys from the defaults? This will make maintenance easier.

beckermr: Or at least we could store them at the top of the file.

esheldon: they are not in the defaults

beckermr: top of the file would be good too.

PR #97, opened 2020-04-22
Realistic bleeds

### LINKED ISSUE #81: Use example star/bleed saturated regions from DC2 like sims
Jim Chiang has kindly generated some examples that we can use.

For a star of a given mag we can look up an example with a similar mag and copy it directly into the mask

### LINKED ISSUE #87: make sat star placement based purely on magnitude
currently one can set a density etc., but remove all those keyword options

esheldon:
Currently just loading all g and using for every band since we only have g band.

Will add other bands when cori is back up

closes #87 
closes #81


esheldon:
unfortunately the code I wrote to extract these from the example images is on my desktop at work, which I can no longer access :(

esheldon:
bleed

descwl_shear_sims/tests/test_gen_star_masks.py
```diff
@@ -1,71 +1,74 @@
...
+            se_obs = data['r'][0]
+            mask = se_obs.bmask.array
+
+            w = np.where((mask & SAT) != 0)
+
+            if trial == 1:
+                nmarked = w[0].size
+            else:
+                assert w[0].size == nmarked
+    except OSError:
```
beckermr, line 72: We should use pytest.mark.skipif for these

beckermr:
This line looks wrong: https://github.com/LSSTDESC/descwl-shear-sims/blob/bleeds/descwl_shear_sims/simple_sim.py#L911

descwl_shear_sims/tests/test_gen_star_masks.py
esheldon: you had implemented it that way in some test I thought. Do you want to do the change?

beckermr: sure!

esheldon:
that bug had not manifested because I had only been using one epoch

esheldon:
0.3 you mean?

beckermr:
Whoops yes.

esheldon:
adaptive-mask

esheldon:
@beckermr are you ok with these new changes?

descwl_shear_sims/render_sim.py
```diff
@@ -46,9 +46,9 @@ def append_wcs_info_and_render_objs_with_psf_shear(
     shear_scene : bool
         If True, the object positions and their shapes are sheared. Otherwise,
         only the object shapes are sheared.
-    expand_star_stamps: bool
-        If True, expand bright star stamps to avoid rendering issues, default
-        True
+    threshold: float
+        For bright stars we will calculate the radius at which we reach
```
beckermr, line 50: Doc string is wrong. I think it does it for all objects.

esheldon: there is a mag cut at 18

esheldon: but that is incorrect grammar :)

esheldon:
I'm gonna do a run at ANL before I merge this.

esheldon:
I think this is as far as I want to go with this PR, I'm gonna  merge and start other branches

PR #100, opened 2020-05-14
Distort wcs

### LINKED ISSUE #48: distortion terms in the WCS

esheldon:
closes #48 

TODO
* add code in the sim to use this wcs
* add more than one example wcs
* keep the header around so we can use it to create the wcs in the lsst code
* make descwl_coadds look for that header and use if it it is available

### DIFF
--- a/descwl_shear_sims/cache_tools.py
+++ b/descwl_shear_sims/cache_tools.py
@@ -5,3 +5,8 @@ import fitsio
 @functools.lru_cache(maxsize=8)
 def cached_catalog_read(fname):
     return fitsio.read(fname)
+
+
+@functools.lru_cache(maxsize=8)
+def cached_header_read(*, fname, ext):
+    return fitsio.read_header(fname, ext=ext)
--- a/descwl_shear_sims/tests/test_gen_sip_wcs.py
+++ b/descwl_shear_sims/tests/test_gen_sip_wcs.py
@@ -1,0 +1,139 @@
+import os
+import pytest
+import numpy as np
+import galsim
+
+from ..gen_sip_wcs import gen_sip_wcs
+
+
+@pytest.mark.skipif(
+    "CATSIM_DIR" not in os.environ,
+    reason='simulation input data is not present')
+def test_gen_sip_wcs_smoke():
+    seed = 42
+    world_origin = galsim.CelestialCoord(
+        ra=20 * galsim.degrees,
+        dec=-10 * galsim.degrees)
+    origin = galsim.PositionD(x=10, y=11)
+    gen_sip_wcs(
+        rng=np.random.RandomState(seed=seed),
+        position_angle_range=(0, 360),
+        dither_range=(-0.5, 0.5),
+        scale=0.25,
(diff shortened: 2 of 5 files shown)


esheldon:
the galsim wcs had the ability to write out the fits hdr keywords to a dict, so we don't need to pass along the header

esheldon:
Use the `wcs._writeHeader` method.

beckermr:
We might want to ask @rmjarvis if we can convert this to a public API so it doesn't change on us.

esheldon:
There is a public one `writeToFitsHeader` but it takes a bounds instance that is not needed

beckermr:
A `writeToDict` method might be useful generally and fit the bill.

esheldon:
looking at the code, I guess I could fake it with `galsim.BoundsI(1, anything, 1, anything) ` since it only checks that the xmin and xmax are 1

esheldon:
that code will go into descwl_coadd, so I'm ready to merge this one if you are matt

descwl_shear_sims/cache_tools.py
```diff
@@ -5,3 +5,8 @@
 @functools.lru_cache(maxsize=8)
 def cached_catalog_read(fname):
     return fitsio.read(fname)
+
+
+@functools.lru_cache(maxsize=8)
```
beckermr, line 10: This max size is really small, especially since we could have a lot images going in eventually. Do we want to increase it?

descwl_shear_sims/tests/test_gen_sip_wcs.py
```diff
@@ -0,0 +1,139 @@
...
+        scale=0.25,
+        scale_frac_std=0.1,
+        shear_std=0.1,
+        world_origin=world_origin,
+        origin=origin)
+
+
+@pytest.mark.skipif(
+    "CATSIM_DIR" not in os.environ,
+    reason='simulation input data is not present')
```
beckermr, line 31: These tests don't appear to depend on the catsim dir

beckermr: I am not sure they are running.

esheldon: the sip version relies on data in the catsim dir

descwl_shear_sims/cache_tools.py
esheldon: that's your original maxsize.  Agreed we might need more, can you open a new issue?

esheldon:
I ran tests locally, true they aren't running in travis since we don't have the catsim data

beckermr:
But the tan-sip doesn't use catsim, so why have the decorators there at all?

descwl_shear_sims/tests/test_gen_sip_wcs.py
beckermr: ahh I see it now. We need this to be controlled by the input RNG eventually.

esheldon:
It reads data from CATSIM_DIR

PR #105, opened 2020-05-26
Send psf dims

esheldon:


esheldon:
I think this one is uncontroversial.

PR #113, opened 2020-06-12
implement rotations, epochs, bands

esheldon:


esheldon:
I ran a test and this works at high s/n

esheldon:
also checked with multiple epochs, looks ok.  So everything is lining up.

beckermr:
What about low s/n?

esheldon:
that will take a day or so, and needs to wait until some other runs finish

esheldon:
(one day if the queues are friendly)

PR #114, opened 2020-06-12
bug not repeating positions exactly

esheldon:


esheldon:
amazing that this worked; metacal hides a lot of things

beckermr:
wow

esheldon:
this was equivalent of having lots of objects very close together, so not that surprising I guess

PR #115, opened 2020-06-24
trivial sim feature parity

esheldon:
Bring trivial sim up to near feature parity with sim

Most features are there, but many of the options are not exposed.  e.g. there is an option to add stars but you don't get to choose how they are added.

The tests all pass but I don't consider that good enough for a merge, partly because more unit tests are needed.  But it is also due to a change in philosophy:

_I now only consider a feature ready if I have run it all the way to recovered shear and I understand the result_

This means waiting until a potentially long test is run, days or even weeks depending on the availability of computing resources.

TODO on this PR

- [x] finish run with stars+wldeblend+PS psf (should finish in a day or two)
- [x] run with star bleeds, cosmics, badcols, etc. (few days to run)
- [x] run with 3 bands
- [ ] add SIP wcs (maybe wait for a separate PR)
- [x] more pixel level unit tests
- [x] normalize use of name bmask rather than mask everywhere in the code
- [ ] consider whether we should try to move the objects using the full wcs and draw in postage stamps (maybe push to separate PR)

### DIFF
--- a/descwl_shear_sims/trivial_sim.py
+++ b/descwl_shear_sims/trivial_sim.py
@@ -28,6 +28,7 @@ DEFAULT_TRIVIAL_SIM_CONFIG = {
     "rotate": False,
     "bands": ["i"],
     "epochs_per_band": 1,
+    "noise_factor": 1.0,
 }
 
 DEFAULT_FIXED_GAL_CONFIG = {
@@ -49,6 +50,7 @@ def make_trivial_sim(
     rotate=False,
     bands=['i'],
     epochs_per_band=1,
+    noise_factor=1.0,
 ):
     """
     Make simulation data
@@ -57,13 +59,26 @@ def make_trivial_sim(
     ----------
     rng: numpy.random.RandomState
         Numpy random state
+    galaxy_catalog: catalog
+        E.g. WLDeblendGalaxyCatalog or FixedGalaxyCatalog
     coadd_dim: int
         Default 351
     g1: float
         Shear g1 for galaxies
     g2: float
         Shear g2 for galaxies
-
+    psf_dim: int, optional
+        Dimensions of psf image.  Default 51
+    dither: bool, optional
+        Whether to dither the images at the pixel level, default False
+    rotate: bool, optional
+        Whether to rotate the images randomly, default False
+    bands: list, optional
+        Default ['i']
+    epochs_per_band: int, optional
+        Number of epochs per band
+    noise_factor: float, optional
+        Factor by which to multiply the noise, default 1
     """
 
     se_dim = (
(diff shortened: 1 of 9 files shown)


esheldon:
At some point we should rename this from trivial to something better.

beckermr:
> At some point we should rename this from trivial to something better.

The only difference is a fully flat geometry. Maybe we could use that?

esheldon:
I'm using TAN wcs which isn't flat, what do you mean?

That does remind me that SIP is not currently available, will add to TODO

beckermr:
Well when you render with galsim, you are using only the WCS jacobian at the center of the field. Then you declare a TAN WCS that has the same jacobian at the center.

beckermr:
The sims I wrote attempt to use the actual WCS everywhere.

descwl_shear_sims/trivial_sim.py
```diff
@@ -250,3 +318,1004 @@ def make_wcs(*, scale, image_origin, world_origin, theta=None):
...
+
+    objects = galsim.Add(convolved_objects)
+
+    # everything gets shifted by the dither offset
+    image = objects.drawImage(
+        nx=dim,
+        ny=dim,
+        wcs=se_wcs,
+        offset=offset,
+    )
```
beckermr, line 582: When you do this, galsim always calls `.local` on the input WCS and then draws with that jacobian.

esheldon:
right.  I think this is demonstrating that for these small fields the jacobian is a good enough approximation.  We don't plan to simulate the whole sky here, just tiny portions

And previous tests showeed that even with typical distortions it is ok

descwl_shear_sims/trivial_sim.py
esheldon: wtf?  what's the point of having complicated wcs support if it is not used?

beckermr:
I am not so sure. We use different bits of the jacobians in different places.

esheldon:
I think this is ready.

I want to put off some additional features for a separate PR, such as optimizations and testing with separate coadds in each band

descwl_shear_sims/trivial_sim.py
```diff
@@ -250,3 +318,1004 @@ def make_wcs(*, scale, image_origin, world_origin, theta=None):
...
+            se_wcs=se_wcs,
+            se_origin=se_origin,
+        )
+        for i, obj in enumerate(bright_convolved_objects):
+
+            obj.drawImage(
+                image=timage,
+                offset=offset,
+                method='phot',
+                n_photons=10_000_000,
```
beckermr, line 619: why photons here?

beckermr: I think we need to compute the number of photons from the exposure time etc so that the noise is right?

esheldon: Using photons avoids fft artifacts on these huge bright stars.

We don't care about getting the noise right, we mask them entirely anyway

esheldon:
by "full module" do you mean sub-package?

beckermr:
Yes

esheldon:
I've added the docs.

I would like to do the refactor of trivial_sim.py into a sub-package in a separate PR

PR #124, opened 2021-01-19
don't rotate the catalog, we do this in the Catalog class

### LINKED ISSUE #119: don't randomly rotate when reading wldeblend catalog
We are already rotating the galsim objects, no need to modify the catalog

esheldon:
closes #119


esheldon:
need to move to something other than travis

PR #133, opened 2021-04-19
Add mockups for DM psf objects

esheldon:


esheldon:
looks like its failing during the stackvana install

esheldon:
@beckermr any ideas how to fix the stackvana install?

beckermr:
I think this should be close to fixed. There is no need for another channel now. It is built on conda-forge.

esheldon:
thanks!

esheldon:
The next step is to replace SEObs with stack exposures.

* image is represented by an exposure (with variance, mask, psf)
* noise is also represented by exposure
* The make_sim would then return separate exposure lists for image and noise image

This is a big change

esheldon:
This reinforces the maxim to focus on data structures and the code will fall in place.

If we had sorted out the DM data structures from the start we would not need all this code restructuring

esheldon:
needed to add stack WCS code before making exposures

esheldon:
I've got a temporary separate code path in the `make_dmsim` function (as opposed to `make_sim`) until I can demonstrate the new version agrees with the old, and have adapted the coadding code to use this version.

esheldon:
currently there are some parallel versions here, e.g. the `make_dmsim` code that will become `make_sim` after more checks.  But those checks must be done separately using the updated coadding code and a full run on a grid sim with PS psf

esheldon:
More testing is needed, but it will  need to be a shear recover test.  we need to merge this to test further

beckermr:
I won't get to this until next week.

esheldon:
no problem

descwl_shear_sims/sim/sim.py
```diff
@@ -334,6 +341,342 @@ def make_seobs(
...
+
+            bdata_list.append({'exp': exp, 'noise_exp': noise_exp})
+
+        band_data[band] = bdata_list
+
+    xoff = 3000
+    coadd_bbox = geom.Box2I(
+        geom.IntervalI(min=xoff + 0, max=xoff + coadd_dim-1),
+        geom.IntervalI(min=0, max=coadd_dim-1),
+    )
```
beckermr, line 475: Why is xoff fixed here?

descwl_shear_sims/sim/sim.py
```diff
@@ -334,6 +341,342 @@ def make_seobs(
...
+            g2=g2,
+        )
+        objlist, shifts, bright_objlist, bright_shifts, bright_mags = get_objlist(
+            galaxy_catalog=galaxy_catalog,
+            survey=survey,
+            g1=g1,
+            g2=g2,
+            star_catalog=star_catalog,
+            noise=noise_per_epoch,
+        )
```
beckermr, line 424: What is going on with these two calls? It appears the first versions of shifts and objlist are being discarded?

descwl_shear_sims/sim/sim.py
```diff
@@ -334,6 +341,342 @@ def make_seobs(
...
+        Dimensions of psf image that will be drawn when psf func is called
+    dither: bool
+        If set to True, dither randomly by a pixel width
+    rotate: bool
+        If set to True, rotate the image randomly
+    cosmic_rays: bool
+        If True, put in cosmic rays
+    bad_columns: bool
+        If True, put in bad columns
+    """
```
beckermr, line 541: Doc string is not complete.

descwl_shear_sims/sim/sim.py
```diff
@@ -334,6 +341,342 @@ def make_seobs(
...
+            image += timage
+
+            calculate_and_add_bright_star_mask(
+                image=timage.array,
+                bmask=bmask.array,
+                shift=bright_shifts[i],
+                wcs=se_wcs,
+                origin=se_origin,
+                threshold=mask_threshold,
+            )
```
beckermr, line 631: A comment on why we photon shoot here would be helpful.

descwl_shear_sims/sim/dmpsfs.py
```diff
@@ -0,0 +1,179 @@
...
+        gspsf: GSObject
+            A galsim GSObject representing the psf
+        psf_dim: int
+            Dimension of the psfs to draw, must be odd
+        wcs: galsim WCS
+            WCS for drawing
+        """
+        ImagePsf.__init__(self)
+
+        if psf_dim // 2 == 0:
```
beckermr, line 53:
```suggestion
        if psf_dim % 2 == 0:
```

descwl_shear_sims/sim/dmpsfs.py
```diff
@@ -0,0 +1,179 @@
...
+        pspsf: PowerSpectrumPSF
+            The power spectrum psf
+        psf_dim: int
+            Dimension of the psfs to draw, must be odd
+        wcs: galsim WCS
+            WCS for drawing
+        """
+        ImagePsf.__init__(self)
+
+        if psf_dim // 2 == 0:
```
beckermr, line 161:
```suggestion
        if psf_dim % 2 == 0:
```

descwl_shear_sims/sim/dmpsfs.py
```diff
@@ -0,0 +1,179 @@
+import galsim
```
beckermr, line 1: The class hierarchy in this file is mangled a bit. The shared methods in the fixed PSF and the power spectrum ones should be put into a new base class that inherits from ImagePsf. Then those two objects should inherit from that new base class.

descwl_shear_sims/tests/test_dmwcs.py
```diff
@@ -0,0 +1,53 @@
...
+    print(skypos.getRa(), skypos.getDec())
+    print(type(skypos.getRa()))
+
+    print()
+
+    gs_skypos = galsim_wcs.toWorld(gs_pos)
+    print(type(gs_skypos))
+    print(gs_skypos)
+    print(gs_skypos.ra, gs_skypos.dec)
+    # stop
```
beckermr, line 42: We should convert these print statements into asserts about the types or simply remove them.

descwl_shear_sims/tests/test_dmsim.py
```diff
@@ -0,0 +1,425 @@
+import os
+import pytest
+import numpy as np
+# import galsim
```
beckermr, line 4:
```suggestion
```

descwl_shear_sims/tests/test_dmsim.py
```diff
@@ -0,0 +1,425 @@
...
+import lsst.afw.image as afw_image
+import lsst.afw.geom as afw_geom
+# import lsst.geom as geom
```
beckermr, line 7:
```suggestion
```

descwl_shear_sims/tests/test_dmsim.py
```diff
@@ -0,0 +1,425 @@
...
+    # y = 10.1
+    # pos = geom.Point2D(x=x, y=y)
+    # gs_pos = galsim.PositionD(x=x, y=y)
+    #
+    # dmpsf = exp.getPsf()
+    # dmpsfim = dmpsf.computeImage(pos)
+    #
+    # gspsf = seobs.get_psf(x, y)
+    #
+    # assert np.all(dmpsfim.array == gspsf.array)
```
beckermr, line 98: Should we fix this bug in this PR? If not, we should make an github issue on it if we have not already.

descwl_shear_sims/tests/test_dmsim.py
```diff
@@ -0,0 +1,425 @@
...
+
+@pytest.mark.parametrize('dither,rotate', [
+    (False, False),
+    (False, True),
+    (True, False),
+    (True, True),
+])
+def test_compare_sim(dither, rotate):
+    """
+    test sim can run
```
beckermr, line 30: description is wrong here

descwl_shear_sims/tests/test_dmsim.py
```diff
@@ -0,0 +1,425 @@
...
+        psf=psf,
+    )
+
+
+@pytest.mark.parametrize(
+    "cosmic_rays, bad_columns",
+    [(True, True),
+     (True, False),
+     (False, True),
+     (True, True)],
```
beckermr, line 300:
```suggestion
     (False, False)],
```

beckermr (changes requested):
Some minor things but looks great!

descwl_shear_sims/sim/sim.py
esheldon: The point was just to have it not start at the zero of the larger reference grid.  I didn't think it mattered for this if we varied it.

If we want to get more realistic we can, making each slice part of a larger "tract" like lsst and having it tile that area.

descwl_shear_sims/sim/sim.py
esheldon: looks like a bug, forgot to remove an old call

descwl_shear_sims/sim/dmpsfs.py
esheldon: whoops

descwl_shear_sims/tests/test_dmsim.py
esheldon: maybe; this was for comparison with the old code, which will be removed.

I think the full tests I'm doing *should* remove the need for that comparison

esheldon:
I'm going to merge this, can you approve @beckermr ?

PR #143, opened 2021-05-26
WIP shear integration tests

beckermr:
This PR has a start for integration tests for shear recovery.

### DIFF
--- a/shear_meas_tests/test_shear_meas.py
+++ b/shear_meas_tests/test_shear_meas.py
@@ -1,0 +1,238 @@
+import time
+import copy
+import numpy as np
+import ngmix
+import tqdm
+import joblib
+
+import pytest
+
+import metadetect.lsst_metadetect as lsst_metadetect
+import descwl_shear_sims as sim
+import descwl_coadd.coadd as coadd
+
+CONFIG = {
+    "model": "wmom",
+    "bmask_flags": 0,
+    "metacal": {
+        "use_noise_image": True,
+        "psf": "fitgauss",
+    },
+    "psf": {
+        "model": "gauss",
+        "lm_pars": {},
+        "ntry": 2,
+    },
+    "weight": {
+        "fwhm": 1.2,
+    },
+    "detect": {
+        "thresh": 10.0,
+    },
+    'meds': {},
+}
+
+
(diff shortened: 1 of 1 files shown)


beckermr:
@esheldon There is a lot going here I don't follow.

1. I am seeing a 30% shear bias when putting this sim through coadding and then metadetect. There must be a bug somewhere.
2. The response to shear in g2 is negative, which is extremely odd.

Can you look through the code?

beckermr:
OK, I figured out the bias. The stack is shredding objects when the S/N is too high. I was using mag=14 in our units and seeing it not find 36 objects every time.  This caused big non-linearities in the response to shear which breaks mdet.

I still think we need to track down the minus sign in R22. I'll make an issue.

beckermr:
This PR is ready for review!

shear_meas_tests/test_shear_meas.py
```diff
@@ -0,0 +1,236 @@
...
+def _boostrap_m_c(pres, mres):
+    m, c = _meas_m_c_cancel(pres, mres)
+    bdata = _bootstrap_stat(pres, mres, _meas_m_c_cancel, 14324, nboot=500)
+    merr, cerr = np.std(bdata, axis=0)
+    return m, merr, c, cerr
+
+
+def _run_sim_one(*, seed, mdet_seed, g1, g2, **kwargs):
+    sim_data = _make_lsst_sim(seed=seed, g1=g1, g2=g2, **kwargs)
+    mbc = coadd.MultiBandCoaddsDM(
```
esheldon, line 125: this class is going to disappear when I finish #146

esheldon (approved):

shear_meas_tests/test_shear_meas.py
esheldon: I will also remove `MultiBandCoadds` and all that will remain is the online coadds (renamed)

beckermr: Ok. We can adjust this code after that pr merges.

esheldon: sorry, I referenced a PR in this repo, I meant https://github.com/LSSTDESC/descwl_coadd/pull/51

PR #146, opened 2021-06-01
Nonoise

esheldon:
removing noise images and old sim codes


descwl_shear_sims/wcs/dmwcs.py
```diff
@@ -59,9 +59,11 @@ def make_dm_wcs(galsim_wcs):
     return stack_wcs
 
 
-def make_coadd_dm_wcs(coadd_origin):
+# def make_coadd_dm_wcs(coadd_origin):
```
beckermr, line 62: remove comment here

setup.py
```diff
@@ -2,7 +2,7 @@
 
 setup(
     name="descwl_shear_sims",
-    version="0.3.0",
+    version="0.3.1",
```
beckermr, line 5: We should read this out of the package. I can do this in another PR.

beckermr (changes requested):
Some small things and then we can merge!

setup.py
esheldon: I'll do it

descwl_shear_sims/wcs/dmwcs.py
esheldon: ok

esheldon:
yes!

beckermr (approved):

PR #149, opened 2021-06-14
add catalog len()

esheldon:


esheldon:
also added star_config entry for sim configs, which is an extra config that can be used for constructing the star catalog but is not otherwise checked, similar to gal_config

beckermr (approved):
What issue/idea motivated this change?

esheldon:
this is to provide information downstream; the number of objects is typically a poisson deviate so it may be useful to extract the actual  number of true objects put in

esheldon:
as for star config, I had forgotten to expose the density= keyword for the StarCatalog downstream, and in doing so realized I also had not allowed the star_config entry to be in the sim config

PR #150, opened 2021-06-24
Se dims and hidden method

esheldon:
Add feature where user can force an exact se_dim (single epoch image dim)

Also add hidden method to DM psfs that gets called

### DIFF
--- a/descwl_shear_sims/psfs/dmpsfs.py
+++ b/descwl_shear_sims/psfs/dmpsfs.py
@@ -78,7 +78,7 @@ class FixedDMPSF(ImagePsf):
 
         return self._make_image(image_pos, offset=offset)
 
-    def computeKernelImage(self, image_pos):  # noqa
+    def computeKernelImage(self, image_pos, color=None):  # noqa
         """
         compute a centered kernel image appropriate for convolution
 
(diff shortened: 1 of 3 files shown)


descwl_shear_sims/psfs/dmpsfs.py
```diff
@@ -78,14 +78,33 @@ def computeImage(self, image_pos):  # noqa
...
+        color: afw_image.Color
+            A color, which is ignored
+        """
+
+        return self._doComputeKernelImage(
+            image_pos=image_pos,
+            color=color,
+        )
+
+    def _doComputeKernelImage(self, image_pos, color=None):  # noqa
```
beckermr, line 98: Is there DM code that accesses hidden methods?

esheldon: Yes :(

beckermr: Ouch. We should file a ticket if you have the details. This is bad form!

esheldon: agreed

esheldon: Its hard to track down because it happens in C++ and the names are different there

esheldon: Here is an official test where they do it

https://github.com/lsst/afw/blob/master/tests/test_psf_trampoline.py#L66

esheldon: If that is a must override it should be part of the public API

esheldon:
Shall we merge anyway?

beckermr (approved):
Yes we should merge!

PR #151, opened 2021-07-12
Implementation of correlated noise

aguinot:
A few changements to implement correlated noise.
The way it is done, the original test should not be affected.
I  have created a new pytest `test_shear_meas_corr_noise.py` where the correlated noise is ON by default (this can probably be adjusted).

As we discuss with @beckermr the flux of the galaxies has to be change in order to see any effect of this. It is not included in this pull request since I did not find a way to do it without hard coded something, yet.

### DIFF
--- a/descwl_shear_sims/sim.py
+++ b/descwl_shear_sims/sim.py
@@ -38,6 +38,7 @@ DEFAULT_SIM_CONFIG = {
 def make_sim(
     *,
     rng,
+    galsim_rng,
     galaxy_catalog,
     coadd_dim,
     g1,
@@ -54,6 +55,9 @@ def make_sim(
     cosmic_rays=False,
     bad_columns=False,
     star_bleeds=False,
+    corr_noise=False,
+    g1_noise=0.01,
+    g2_noise=0.01,
 ):
     """
     Make simulation data
@@ -62,6 +66,8 @@ def make_sim(
     ----------
     rng: numpy.random.RandomState
         Numpy random state
+    galsim_rng: galsim.BaseDeviate
+        Galsim random state (used for noise realisation)
     galaxy_catalog: catalog
         E.g. WLDeblendGalaxyCatalog or FixedGalaxyCatalog
     coadd_dim: int
@@ -94,6 +100,12 @@ def make_sim(
         If True, add cosmic rays
     bad_columns: bool
         If True, add bad columns
+    corr_noise: bool
+        If True, make correlated noise
+    g1_noise: float
+        g1 for shear correlated noise
+    g2_noise: float
+        g2 for shear correlated noise
     """
 
     coadd_wcs, coadd_bbox = make_coadd_dm_wcs(coadd_dim)
@@ -132,6 +144,7 @@ def make_sim(
         for epoch in range(epochs_per_band):
             exp = make_exp(
                 rng=rng,
+                galsim_rng=galsim_rng,
                 band=band,
                 noise=noise_per_epoch,
                 objlist=objlist,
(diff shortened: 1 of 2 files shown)


descwl_shear_sims/sim.py
```diff
@@ -55,6 +54,7 @@ def make_sim(
     cosmic_rays=False,
     bad_columns=False,
     star_bleeds=False,
+    galsim_rng=None,
```
beckermr, line 57: Setting this to `None` will cause the seed to differ each time and make the sim not reproducible. Instead, we need to draw a seed from the input numpy RNG and instantiate the galsim_rng internally.

descwl_shear_sims/sim.py
```diff
@@ -197,7 +197,6 @@ def make_sim(
 def make_exp(
```
beckermr: I'd move all of the galsim RNG instantiations to this function. This keeps the APIs the same and properly decouples the RNG from the rest of the code.

descwl_shear_sims/sim.py
aguinot: That is how it is suppose to be done. The thing is, it is not used by the original script, so to not change it I did this "trick" to avoid having an error. What I would suggest is to also use the galsim noise generator even on the original test so this can be propagated correctly. That would mean changing [this line](https://github.com/LSSTDESC/descwl-shear-sims/blob/02fd39dbe4c982c575193ea861bbaa51b99466f0/descwl_shear_sims/sim.py#L333).

descwl_shear_sims/sim.py
aguinot: Wouldn't it be better to have it declare at the same time as the numpy rng?

descwl_shear_sims/sim.py
beckermr: I'm not being clear. We don't use the galsim RNG infrastructure in this sim package. So the RNG instantiation has to be changed. We cannot merge this code in its current form.

descwl_shear_sims/sim.py
beckermr: Nope. The sim seed is set by the numpy RNG seed. We need to chain the galsim one to it via what I said.

descwl_shear_sims/sim.py
aguinot: What I mean is that this is not use by the original purest script and it is correctly instantiate in the one I have done where it is used so normally it should not be a problem. But I agree on the fact that this looks ugly at the moment.

descwl_shear_sims/sim.py
aguinot: Ok I think I understand better.
So I should do something like [this](https://github.com/LSSTDESC/descwl-shear-sims/blob/02fd39dbe4c982c575193ea861bbaa51b99466f0/shear_meas_tests/test_shear_meas_corr_noise.py#L191) `rng.randint(low=1, high=2**29, size=ntrial)` ?

beckermr: Yeah just set the size to 1 and feed this as a seed to the GALSIM RNG

shear_meas_tests/test_shear_meas_corr_noise.py
```diff
@@ -0,0 +1,258 @@
...
+        "lm_pars": {},
+        "ntry": 2,
+    },
+    "weight": {
+        "fwhm": 1.2,
+    },
+    "detect": {
+        "thresh": 10.0,
+    },
+    'meds': {},
```
beckermr, line 33:
```suggestion
    'meds': {},
    'subtract_sky': False,
```

descwl_shear_sims/sim.py
```diff
@@ -221,6 +233,10 @@ def make_exp(
     star_bleeds=False,
     sky_n_sigma=None,
     draw_method='auto',
+    galsim_rng=None,
```
beckermr, line 236: We do not need this.

aguinot: You right, I forgot to remove the keyword. It should not be used in the code.

Issue #117, opened 2020-07-13
Add optimized drawing


esheldon:
drawing in small postage stamps and adding to the image is much faster but we need to ensure we can still recover the shear, in other words there are no simulation artifacts associated with this

esheldon:
There is a nice way to do this with newer galsim
```python
image = ImageF(nx,ny, wcs=complex_wcs)
shear = Shear(g1=g1,g2=g2)
for obj, world_pos in zip(obj_list, world_pos_list):
    image_pos = (image.wcs.toImage(world_pos) - image.true_center).shear(shear) + image.true_center
    stamp = obj.shear(shear).drawImage(wcs=image.wcs, center=image_pos)
    b = stamp.bounds & image.bounds
    image[b] += stamp[b]
```
we currently calculated the offsets rather than getting world pos and then converting to offsets.  Not sure which I would prefer in the long run, but the later is certainly easier if we have some full sky catalog.

Note I'm not sure if this will always work as well as the "Add all objects then draw" version.  The latter adds all the objects together so it has a good idea of how bright each part of the image is.  e.g. the value in a pixel is the sum over all objects and can be a lot higher than any individual object.  Without that information the drawing might not set appropriate thresholds and, for example, make the stamp too small which could cause artifacts that will fool the shear code

esheldon:
Also not that the shearing of the image position is not actually implemented yet :)

rmjarvis:
FYI, shearing positions is implemented now in GalSim v2.3.
Syntax is:
```
sheared_pos = pos.shear(shear)
```
cf. http://galsim-developers.github.io/GalSim/_build/html/pos.html

esheldon:
How do I achieve this if I have a set of offsets in arcseconds rather than a set of world positions?  Do I need to do the calculations by hand or is there a way to do this using galsim primitives?

beckermr:
You want pos to be the distance to the center of the uv plane in arcseconds.

esheldon:
my shifts are in uv.

To use mikes' formulat I need to transform uv to pixels.  Just wondering if there was built in way to do it or do I need to type out the linear transformation

esheldon:
I think it would be this in uv
```python
sheared_uv_shift = uv_shift.shear(shear)
pixel_shift = (transformation here)
image_pos = image_center + pixel_shift
```

beckermr:
ahhhhh - this likely depends on how the coadd wcs is represented.

[...]

PR #153, opened 2021-07-29
make galsim draw method optional, default auto

esheldon:


esheldon:
I think just turning on 'phot' is correct because wldeblend gives flux in electrons

However, it can use a ton of memory so probably want to put a cap on number of photons

esheldon:
This gets pretty slow in some cases if there are objects with a lot of photons.

Because we are drawing all objects at once (other than BRIGHT) we can't control this by using fewer photons in some cases.

esheldon:
I think we would need to implement #117 to make this faster

esheldon:
@rmjarvis does not think we can speed up the photon shooting, we making it optional and we can do targeted studies

esheldon:
@beckermr this one should be uncontroversial

beckermr (approved):

PR #154, opened 2021-08-02
Optimize drawing

esheldon:


descwl_shear_sims/objlists.py
```diff
@@ -41,94 +29,21 @@ def get_objlist(*, galaxy_catalog, survey, g1, g2, star_catalog=None, noise=None
         )
         sobjlist, sshifts, bright_objlist, bright_shifts, bright_mags = res
 
-        objlist = objlist + sobjlist
-
-        shifts = np.hstack((shifts, sshifts))
+        # objlist = objlist + sobjlist
+        # shifts = np.hstack((shifts, sshifts))
```
beckermr, line 33: Remove this code if not needed.

descwl_shear_sims/stars.py
```diff
@@ -77,34 +77,31 @@ def get_objlist(self, *, survey, noise):
 
         Returns
         -------
-        [galsim objects]
+        [galsim objects], [shifts],
+        [bright_objlist], [bright_shifts], [bright_mags]
```
beckermr, line 81: doc string wrong?

descwl_shear_sims/sim.py
```diff
@@ -390,6 +360,163 @@ def make_exp(
...
+        convolved_object = get_convolved_object(obj, psf, image_pos)
+
+        maxn = 10_000_000
+        n_photons = int(obj.flux) if obj.flux < maxn else maxn
+
+        stamp = convolved_object.drawImage(
+            center=image_pos, wcs=local_wcs,
+            method='phot',
+            n_photons=n_photons,
+            maxN=1_000_000,
```
beckermr, line 458: Won't this override n_photons for things between 1 million and 10 million?

beckermr (changes requested):
A few questions. We need to run a shear recovery test at low S/N before we merge this one.

descwl_shear_sims/sim.py
esheldon: What galsim does when you send n_photons when it does not match flux is to put extra counts into each photon.

esheldon: maxN means shot at most that many at a time; do batches of 1_000_000 rather than all at once, to save memory

beckermr: Then why are you adjusting n_photons with the if statement right before this?

beckermr: ohhhhhhhhhhh. I get it now. LGTM!

esheldon: Up to 10 million I shoot photons 1-1 with the flux, above that I shoot fewer to save time

descwl_shear_sims/stars.py
esheldon: Looks right to me

beckermr: ohh woops I only read the last line! sorry!

descwl_shear_sims/objlists.py
esheldon: fixed

beckermr (approved):

esheldon:
With this speedup the coadding now dominates the runtime.  Therefore it would be good for the slower tests to be able to run without coadding.  I'll be working to support that in the coadds repo before I begin a long run

Possible tests

- [ ] grid test at high s/n
- [ ] grid test high s/n with stars
- [ ] grid test low s/n with stars
- [ ] full test with wldeblend and stars

rmjarvis (approved):
Looks good.  Just a few misc comments.

descwl_shear_sims/sim.py
```diff
@@ -390,6 +360,163 @@ def make_exp(
...
+            shift.y * galsim.arcsec,
+        )
+
+        image_pos = wcs.toImage(world_pos)
+        local_wcs = wcs.local(image_pos=image_pos)
+
+        convolved_object = get_convolved_object(obj, psf, image_pos)
+
+        maxn = 10_000_000
+        n_photons = int(obj.flux) if obj.flux < maxn else maxn
```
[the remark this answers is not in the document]
esheldon: Thanks for the tip, I'll implement this

beckermr:
We really should do the low s/n test without stars.

descwl_shear_sims/sim.py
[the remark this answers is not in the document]
esheldon: It looks like you have to give `0` rather than `None`
```python
        # Make sure the type of n_photons is correct and has a valid value:
>       if n_photons < 0.:
E       TypeError: '<' not supported between instances of 'NoneType' and 'float'
``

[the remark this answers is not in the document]
rmjarvis: Sorry, yes.  The default is 0, not None.  My mistake.

Issue #155, opened 2021-08-09
always dither objects


esheldon:
Objects should be dithered even if the image is not

I ran into this with grid tests without image dithering.

esheldon:
hmm, just checking now and this is happening.

But I saw a clear difference in certain cases with and without image dithering

PR #159, opened 2021-09-17
add pair layout and FixedPairGalaxyCatalog

esheldon:
pair sims

### DIFF
--- a/descwl_shear_sims/shifts.py
+++ b/descwl_shear_sims/shifts.py
@@ -167,7 +167,7 @@ def get_pair_shifts(*, rng, sep):
 
     shifts = np.zeros(2, dtype=[('dx', 'f8'), ('dy', 'f8')])
 
-    angle = rng.uniform(low=0, high=2*np.pi)
+    angle = rng.uniform(low=0, high=np.pi)
     shift_radius = sep / 2
 
     dx = np.cos(angle)*shift_radius
--- a/descwl_shear_sims/shifts.py
+++ b/descwl_shear_sims/shifts.py
@@ -170,13 +170,23 @@ def get_pair_shifts(*, rng, sep):
     angle = rng.uniform(low=0, high=np.pi)
     shift_radius = sep / 2
 
-    dx = np.cos(angle)*shift_radius
-    dy = np.sin(angle)*shift_radius
+    xdither, ydither = SCALE*rng.uniform(low=-0.5, high=0.5, size=2)
 
-    shifts['dx'][0] = dx
-    shifts['dy'][0] = dy
+    dx1 = np.cos(angle)*shift_radius
+    dy1 = np.sin(angle)*shift_radius
+    dx2 = -dx1
+    dy2 = -dy1
 
-    shifts['dx'][1] = -dx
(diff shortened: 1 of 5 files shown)


descwl_shear_sims/shifts.py
```diff
@@ -129,3 +145,38 @@ def get_random_shifts(*, rng, dim, buff, size):
...
+    Returns
+    -------
+    shifts: array
+        Array with dx, dy offset fields for each point, in
+        arcsec
+    """
+
+    shifts = np.zeros(2, dtype=[('dx', 'f8'), ('dy', 'f8')])
+
+    angle = rng.uniform(low=0, high=2*np.pi)
```
beckermr, line 170: This angle should only be to pi and not 2pi.

descwl_shear_sims/shifts.py
```diff
@@ -129,3 +145,38 @@ def get_random_shifts(*, rng, dim, buff, size):
...
+    shift_radius = sep / 2
+
+    dx = np.cos(angle)*shift_radius
+    dy = np.sin(angle)*shift_radius
+
+    shifts['dx'][0] = dx
+    shifts['dy'][0] = dy
+
+    shifts['dx'][1] = -dx
+    shifts['dy'][1] = -dy
```
beckermr, line 180: We should dither these positions?

descwl_shear_sims/shifts.py
esheldon: true, but it will give the same result

descwl_shear_sims/shifts.py
esheldon: We have not been dithering the grid, relying on dithering at the image level.  I actually have an issue on this #155

descwl_shear_sims/shifts.py
beckermr: My student and I found weird results using 2pi but maybe that was another bug.

descwl_shear_sims/shifts.py
beckermr: I think we want dithers in the coadd system too just be safe. It certainly can't hurt. I'd strongly suggest doing it.

esheldon: the shifts codes do not know about the pixel scale, so this would require some restructuring

esheldon: An easier option would be to require image dithers

beckermr: Image dithers still leave the objects at the same spot in the coadd system. This is still a potential source of bias.

esheldon: actually we are dithering the grid, so that issue must be out of date

beckermr:
One more thing. I think the dithers of the two objects should be independent.

esheldon:
how big?  A pixel is significant for these object sizes

esheldon:
I kept sep same because a pixel sized dither would dramatically change the blending

beckermr:
Hmmmm ok. Let's keep it as is then. Thanks for the back and forth here!

PR #160, opened 2021-10-15
Add bulge+disk (bd) and bulge+disk+knots (bdk)

esheldon:
I've added these models to FixedGalaxyCatalog and FixedPairGalaxyCatalog to provide some non-elliptical and non-sersic shapes for testing the deblending.

The bulge and disk are offset randomly, and the bulge is rotated randomly relative to the disk.

The knots are limited in number and fraction of the disk light, and are more compact, drawing from a smaller hlr and a gaussian rather than exponential profile.  This helps avoid knots getting detected separately, although we can loosen that if we want.

### DIFF
--- a/shear_meas_tests/test_shear_meas.py
+++ b/shear_meas_tests/test_shear_meas.py
@@ -7,7 +7,7 @@ import joblib
 
 import pytest
 
-import metadetect.lsst_metadetect as lsst_metadetect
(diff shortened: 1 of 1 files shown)


esheldon:
Hmmm.... the bd and bdk are not really "fixed galaxies" any more.  Maybe should consider a new class for objects that vary.

esheldon:
No hurry on this one @beckermr

PR #162, opened 2021-10-19
try to fix psf bboxes

esheldon:
this was reverse engineered from FixedKernel psfs, I could
not find the c++ code for this


esheldon:
This fixes detection/deblending problems that I  had seen when the bbox wasn't adjusted for object position

PR #180, opened 2021-12-13
ENH add hex layout; fix grid buff

beckermr:
closes #180 and #181 

This PR adds a randomly rotated hexagonal grid to the code.

### DIFF
--- a/descwl_shear_sims/shifts.py
+++ b/descwl_shear_sims/shifts.py
@@ -27,7 +27,7 @@ def get_shifts(
         Buffer region where no objects will be drawn.  Ignored
         for layout 'grid'.  Default 0.
     layout: string
-        'grid' or 'random'
+        'grid', 'pair', 'hex', or 'random'
     nobj: int, optional
         Optional number of objects to draw, defaults to None
         in which case a poission deviate is draw according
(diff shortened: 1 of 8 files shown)


descwl_shear_sims/shifts.py
```diff
@@ -66,12 +66,79 @@ def get_shifts(
                 buff=buff,
                 size=nobj,
             )
+        elif layout == 'hex':
+            shifts = get_hex_shifts(
+                rng=rng,
+                dim=coadd_dim,
+                buff=buff,
+                n_on_side=GRID_N_ON_SIDE,
```
esheldon, line 74: The stack is finding these to be blended at this density, but I think we want them to be not deblended by default.

esheldon: using `n_on_side` 5 instead of 6 seems to work when the objects are s/n = 50

however for the default higher s/n the entire thing gets called one footprint.  So I think this needs some tuning

esheldon (changes requested):

descwl_shear_sims/shifts.py
beckermr: Does it work ok for 4?

esheldon: for 4 it still puts everything in one big footprint

beckermr: Wow really. Idk what to do then. This really is a bug in the stack code.

esheldon:
Another issue is that objects in the hex grid are getting closer than buff to the edge

esheldon:
Hmm... looks like this is also true for the regular grid, looks like a bug

descwl_shear_sims/shifts.py
```diff
@@ -66,12 +66,79 @@ def get_shifts(
...
+    hg, _ = create_hex_grid(nx=nx, ny=ny, rotate_deg=rng.uniform() * 360)
+
+    # convert the spacing to right number of pixels
+    # we also recenter the grid since it comes out centered at 0,0
+    hg *= delta
+    upos = hg[:, 0].ravel()
+    vpos = hg[:, 1].ravel()
+
+    # dither
+    upos += rng.uniform(low=-0.5, high=0.5, size=upos.shape[0])
```
esheldon, line 121: these dithers need to be in arcsec

esheldon:
Matt do you want to have a look?

descwl_shear_sims/shifts.py
```diff
@@ -66,13 +68,81 @@ def get_shifts(
...
+    Parameters
+    ----------
+    rng: numpy.random.RandomState
+        The random number generator
+    dim: int
+        Dimensions of the final image
+    buff: int, optional
+        Buffer region where no objects will be drawn.
+    n_on_side: int
+        Number of objects on each side
```
beckermr, line 97: doc string needs update

beckermr (commented):
one minor thing - nice work!

beckermr:
LGTM! I cannot approve since this is my PR.

PR #187, opened 2022-10-14
Ring test, correct image center, enable write PSF object to pickle, record flux zero point in output exposure

mr-superonion:
1. Enable ring test to reduce error on additive bias (4x 45 degree pairs);
2. Enable dumping PSF object to disk (for reuse);
3. record magnitude zero point in the output exposures;
4. The center of the image was at ndim/2-1 for an image with indexing 1,2..ndim (that is the Gaism indexing). Now fix it to ndim/2+1. For example, a length 3 array, with indexing 1,2,3, the center should be at 2=(3+1)/2.

### DIFF
--- a/descwl_shear_sims/galaxies.py
+++ b/descwl_shear_sims/galaxies.py
@@ -602,7 +602,7 @@ class WLDeblendGalaxyCatalog(object):
             else:
                 area = ((coadd_dim - 2*buff)*SCALE/60)**2
 
-        elif layout == 'random_circle':
+        elif layout == 'random_disk':
             # this layout is random in a circle
             if (coadd_dim - 2*buff) < 2:
                 warnings.warn("dim - 2*buff <= 2, force it to 2.")
@@ -613,7 +613,7 @@ class WLDeblendGalaxyCatalog(object):
                 area = np.pi*radius**2
             del radius
         else:
-            raise ValueError("layout can only be 'random' or 'random_circle' \
+            raise ValueError("layout can only be 'random' or 'random_disk' \
                     for wldeblend")
 
         # a least 1 expected galaxy (used for simple tests)
--- a/descwl_shear_sims/galaxies.py
+++ b/descwl_shear_sims/galaxies.py
@@ -75,10 +75,15 @@ def make_galaxy_catalog(
             )
 
         if gal_type == 'wldeblend':
+            assert layout is not None, \
+                    "Cannot make wldebend catalog with layout=None, please use \
+                    layout='random' or 'random_disk'."
+
             galaxy_catalog = WLDeblendGalaxyCatalog(
                 rng=rng,
                 coadd_dim=coadd_dim,
                 buff=buff,
+                layout=layout,
             )
         elif gal_type in ['fixed', 'varying', 'exp']:  # TODO remove exp
             if layout is None:
--- a/descwl_shear_sims/galaxies.py
+++ b/descwl_shear_sims/galaxies.py
@@ -1,3 +1,4 @@
+import warnings
 import numpy as np
 import os
 import copy
@@ -595,17 +596,27 @@ class WLDeblendGalaxyCatalog(object):
         gal_dens = self._wldeblend_cat.size / (60 * 60)
         if layout=='random':
             # this layout is random in a square
-            area = ((coadd_dim - 2*buff)*SCALE/60)**2
+            if (coadd_dim - 2*buff)<2:
+                warnings.warn("dim - 2*buff <= 2, force it to 2.")
+                area    =   (2**SCALE/60)**2.
+            else:
+                area    =   ((coadd_dim - 2*buff)*SCALE/60)**2
+
         elif layout=='random_circle':
             # this layout is random in a circle
-            radius=(coadd_dim/2. - buff)*SCALE/60
-            area = np.pi*radius**2
+            if (coadd_dim - 2*buff)<2:
+                warnings.warn("dim - 2*buff <= 2, force it to 2.")
+                radius  =   2.*SCALE/60
+                area    =   np.pi*radius**2
+            else:
+                radius=(coadd_dim/2. - buff)*SCALE/60
+                area = np.pi*radius**2
             del radius
         else:
             raise ValueError("layout can only be 'random' or 'random_circle' for wldeblend")
-        # print(area)
 
-        nobj_mean = area * gal_dens
+        nobj_mean = max(area * gal_dens,1) # a least 1 expected galaxy (used for simple tests)
+
         nobj = rng.poisson(nobj_mean)
 
         self.shifts_array = get_shifts(
@@ -615,8 +626,6 @@ class WLDeblendGalaxyCatalog(object):
             layout=layout,
             nobj=nobj,
         )
-        # print(np.min(self.shifts_array['dx'])/SCALE,np.max(self.shifts_array['dx'])/SCALE)
-        # print(np.min(self.shifts_array['dy'])/SCALE,np.max(self.shifts_array['dy'])/SCALE)
 
         num = len(self)
         self.indices = self.rng.randint(
--- a/descwl_shear_sims/galaxies.py
+++ b/descwl_shear_sims/galaxies.py
@@ -76,8 +76,8 @@ def make_galaxy_catalog(
 
         if gal_type == 'wldeblend':
             assert layout is not None, \
-                    "Cannot make wldebend catalog with layout=None, please use \
-                    layout='random' or 'random_disk'."
+                "Cannot make wldebend catalog with layout=None, please use \
+                layout='random' or 'random_disk'."
 
             galaxy_catalog = WLDeblendGalaxyCatalog(
                 rng=rng,
--- a/descwl_shear_sims/galaxies.py
+++ b/descwl_shear_sims/galaxies.py
@@ -76,10 +76,6 @@ def make_galaxy_catalog(
 
         if gal_type == 'wldeblend':
             if layout is None:
-                warnings.warn(
-                    "The input layout is None! Force layout to 'random' for \
-                    galaxy_type=wldeblend"
-                )
                 layout = "random"
 
             galaxy_catalog = WLDeblendGalaxyCatalog(
--- a/descwl_shear_sims/galaxies.py
+++ b/descwl_shear_sims/galaxies.py
@@ -48,45 +48,43 @@ def make_galaxy_catalog(
     sep: float, optional
         Separation of pair in arcsec for layout='pair'
     """
-    if layout == 'pair':
+    if layout == "pair":
         if sep is None:
-            raise ValueError(
-                f'send sep= for gal_type {gal_type} and layout {layout}'
-            )
+            raise ValueError(f"send sep= for gal_type {gal_type} and layout {layout}")
         gal_config = get_fixed_gal_config(config=gal_config)
 
-        if gal_type in ['fixed', 'exp']:  # TODO remove exp
+        if gal_type in ["fixed", "exp"]:  # TODO remove exp
             cls = FixedPairGalaxyCatalog
         else:
             cls = PairGalaxyCatalog
 
         galaxy_catalog = cls(
             rng=rng,
-            mag=gal_config['mag'],
-            hlr=gal_config['hlr'],
-            morph=gal_config['morph'],
+            mag=gal_config["mag"],
+            hlr=gal_config["hlr"],
+            morph=gal_config["morph"],
             sep=sep,
         )
 
     else:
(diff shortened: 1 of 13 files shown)


esheldon:
Note rather than ring test we use the noise cancelling method of Pujols. Note in the mdet-lsst-sim repo I use both a +g1 and -g1 and store both in the output files

mr-superonion:
Yes! I think it is better to use both together! (and this is normally what I did) since ring test is very efficient in reducing error on additive bias.

descwl_shear_sims/sim.py
```diff
@@ -319,7 +331,9 @@ def make_exp(
 
     shear = galsim.Shear(g1=g1, g2=g2)
     dims = [dim]*2
-    cen = (np.array(dims)-1)/2
+    # I think Galsim uses 1 offset. An array with length=dim=5
+    # The center is at 3=(5+1)/2
+    cen = (np.array(dims)+1)/2
```
beckermr, line 336: Galsim is 1-offset by defaulkt but the Rubin science pipelines use 0-offset coordinates. Given how the center is used below I am pretty sure 0-offset is correct here.

descwl_shear_sims/sim.py
```diff
@@ -696,3 +723,20 @@ def get_bright_info_struct():
...
+        pos (PositionD):a galsim position
+        theta (float):  rotation angle [rads]
+    Returns:
+        x2 (ndarray):   rotated coordiantes [x]
+        y2 (ndarray):   rotated coordiantes [y]
+    '''
+    x = pos.x
+    y = pos.y
+    x2 = np.cos(theta)*x-np.sin(theta)*y
+    y2 = np.sin(theta)*x+np.cos(theta)*y
```
beckermr, line 741:
```suggestion
    cost = np.cos(theta)
    sint = np.sin(theta)
    x2 = cost*x - sint*y
    y2 = sint*x + cost*y
```

beckermr (commented):
Thanks for the PR! A few initial comments. I'll need to read more later too.

descwl_shear_sims/sim.py
mr-superonion: Let's look into the details of the original code:
https://github.com/LSSTDESC/descwl-shear-sims/blob/5a693fd926f5e99ea1a0f76857dc66085e313ad5/descwl_shear_sims/sim.py#L322-L324
`cen` defined here adopts DM convention!
 
However, you can see the `cen` is first passed into Galsim and used to defined Galsim WCS:

https://github.com/LSSTDESC/descwl-shear-sims/blob/5a693fd926f5e99ea1a0f76857dc66085e313ad5/descwl_shear_sims/sim.py#L336-L342

After that, the Galsim WCS is used to define a DM WCS 
https://github.com/LSSTDESC/descwl-shear-sims/blob/5a693fd926f5e99ea1a0f76857dc66085e313ad5/descwl_shear_sims/sim.py#L395

and in the function here:
https://github.com/LSSTDESC/descwl-shear-sims/blob/5a693fd926f5e99ea1a0f76857dc66085e313ad5/descwl_shear_sims/wcs/dmwcs.py#L25-L29
The Galsim center is transformed to the DM center by doing ` Point2D(crpix[0]-1, crpix[1]-1)`. 
The problem is that at the beginning, the Galsim center does not use Galsim convention..

mr-superonion: Let's assume the `dim=33`, then we have `cen=16`
The Gaism image should have grids at 1,2,3...33. 
The correct Galsim Center should be at (17,17), however we will get an inconsistent one which is (16,16) 
Then, if we pass this inconsistent Galsim WCS to the function `make_dm_wcs()`, we will get 
`stack_crpix=Point2D(15,15)`, which is not correct.

esheldon: @rmjarvis Do what do you think of this, did we mess up the galsim WCS?

rmjarvis: I don't know.  I would write a unit test that checks that the connection between some particular pixel and its Ra/Dec is right.  I'm not following enough of what you are doing here to know whether you did everything right.  But presumably making a test like that should be straightforward.

mr-superonion: OK!! I will add a really simple test for one galaxy and force the offset to [0,0] to show that for the original code, the galaxy is not at the center of the image but has a (-1,-1) offset to the image center. I have done a similar test (but more complicated) in my shear estimation repo.. but I will put that one here..

beckermr: We have a test like this where we coadd psfs. Maybe we don't check the offset there though.

esheldon:
We need to run a full sim test to show this didn't introduce any shear biases.

esheldon:
Please don't lint with black.  It makes pointless changes (e.g. single to double quotes) and creates conflicts with the flake8 we use for CI

mr-superonion:
> Please don't lint with black. It makes pointless changes (e.g. single to double quotes) and creates conflicts with the flake8 we use for CI

wow, what is the software this repo used for auto-linting? Please let me what is the standard?

esheldon:
I don't auto-format my code, I run flake8 within the editor and make the changes by hand.

mr-superonion:
> We need to run a full sim test to show this didn't introduce any shear biases.

Yes!!! Let me make a test for shear estimation with 4 noiseless galaxies (45 degree rotated; with center offset; but fix the center of the measurement to the center of distortion/ rotation)

esheldon:
This might be covered by the CI with shear_meas_tests subdir

mr-superonion:
Note, the shear estimation test also make sure that the distortion/ rotation center is located at the center of the image array.

beckermr:
What is biased?

mr-superonion:
shear is biased if the measurement center is not the same as the center of the distortion/ rotation for this test!

beckermr:
We've done tons and tons of simulations without this PR and found no shear bias. Under what conditions are you talking about? Are you not detecting objects?

mr-superonion:
This is a special case, since I am using only 4 noiseless galaxies for ring test (each galaxy has an offset and then rotated by 45 degree wrt image center) and run metacal (no detection is run). 
This is to make sure that using 4 galaxies successfully removes shape noise.
Since no detection is run, and each galaxy has center offset from the image center, in the measurement we must force the center of the measurement to the center of the rotation, otherwise we will find a bias but this is because that the ring test does not fully remove shape noise.

descwl_shear_sims/shifts.py
```diff
@@ -197,7 +215,7 @@ def get_grid_shifts(*, rng, dim, buff, spacing):
 
 def get_random_shifts(*, rng, dim, buff, size):
     """
-    get a set of gridded shifts, with random shifts at the pixel scale
+    get a set of gridded shifts in a square, with random shifts at the pixel scale
```
esheldon, line 218: Actually the original doc string was quite wrong, this is  not gridded.

It should be something like "get a set of random shifts in a square"

descwl_shear_sims/shifts.py
```diff
@@ -218,6 +236,11 @@ def get_random_shifts(*, rng, dim, buff, size):
     """
 
     halfwidth = (dim - 2*buff)/2.0
+    if halfwidth < MIN_R_SHIFT:
```
esheldon, line 239: Is the idea that the user might use a buffer that is too  large?

esheldon: Maybe we should actually raise an exception if `dim - 2*buff < 0`

descwl_shear_sims/shifts.py
```diff
@@ -230,6 +253,46 @@ def get_random_shifts(*, rng, dim, buff, size):
     return shifts
 
 
+def get_random_circle_shifts(*, rng, dim, buff, size):
+    """Gets a set of gridded shifts in a circle, with random shifts at the
```
esheldon, line 257: not gridded right?  Maybe "get a set of random shifts on a disk"?

descwl_shear_sims/shifts.py
```diff
@@ -230,6 +253,46 @@ def get_random_shifts(*, rng, dim, buff, size):
     return shifts
 
 
+def get_random_circle_shifts(*, rng, dim, buff, size):
```
esheldon, line 256: It fills a disk, not a circle right?

descwl_shear_sims/sim.py
```diff
@@ -405,6 +420,11 @@ def make_exp(
 
     exp = afw_image.ExposureF(masked_image)
 
+    # Prepare the frc
```
esheldon, line 423: What does this do?

descwl_shear_sims/sim.py
```diff
@@ -115,6 +117,8 @@ def make_sim(
     draw_method: string
         Draw method for galsim objects, default 'auto'.  Set to
         'phot' to get poisson noise.  Note this is much slower.
+    theta0: float
+        The rotation angle of the exposure, in units of radians
```
esheldon, line 121: Should be "The rotation angle of the objects, in units of radians" rather than exposure

descwl_shear_sims/sim.py
```diff
@@ -305,7 +315,9 @@ def make_exp(
     draw_method: string
         Draw method for galsim objects, default 'auto'.  Set to
         'phot' to get poisson noise.  Note this is much slower.
-
+    theta0: float
+        rotation angle of intrinsic galaxies and positions [for ring test],
```
esheldon, line 319: Probably should use this same doc string above

descwl_shear_sims/sim.py
mr-superonion: https://hsc.mtk.nao.ac.jp/pipedoc/pipedoc_4_e/e_tool/index.html
See the last row of  table 1 in the link above.
In HSC pipeline, (I do not know why) the photometric calibration file is called frc...

descwl_shear_sims/sim.py
mr-superonion: done!

descwl_shear_sims/shifts.py
mr-superonion: Done!!

descwl_shear_sims/shifts.py
mr-superonion: Done!!

descwl_shear_sims/shifts.py
mr-superonion: Done

descwl_shear_sims/sim.py
esheldon: I mean, what do these operations do?  Do they set some metadata in the exposure?  How is it retrieved etc.

esheldon: Thanks.  Do you mind putting that as a comment in the code for future reference?

mr-superonion: Done!!

mr-superonion:
> Where do we stand on this PR?

Sorry I have been busy with some other projects. 
I am waiting for the review from you and Matthew!!

esheldon (approved):

beckermr (approved):
There is an API change here which may need to be propagated into downstream users of the code.

esheldon:
can you point that out?

descwl_shear_sims/galaxies.py
```diff
@@ -582,24 +583,49 @@ class WLDeblendGalaxyCatalog(object):
     buff: int, optional
         Buffer region with no objects, on all sides of image.  Ingored
         for layout 'grid'.  Default 0.
+    layout: str, optional
+
     """
-    def __init__(self, *, rng, coadd_dim, buff=0):
+    def __init__(self, *, rng, coadd_dim, buff=0, layout='random'):
```
beckermr, line 589: Here @esheldon. If you are trying to do a new `random_ring` simulation, you need to remember to pass this keyword here in addition to the other places `layout` is used.

beckermr:
Right. Not a breaking change but could be broken in some cases.

esheldon:
@mr-superonion can you pass on that keyword indicated above?

mr-superonion:
I think now it is consistent!!

descwl_shear_sims/galaxies.py
```diff
@@ -74,10 +75,15 @@ def make_galaxy_catalog(
             )
 
         if gal_type == 'wldeblend':
+            assert layout is not None, \
```
esheldon, line 78: In the past the default has been random, so we should also default to that here rather than making it a required argument

esheldon (changes requested):

esheldon (approved):

esheldon:
Thanks for this PR, I think this will be very useful.

PR #196, opened 2023-07-19
update flags name for test

esheldon:


esheldon:
the updated test only runs on push to master

Issue #197, opened 2023-07-21
add cluster shears


lbaumo:
you can use clmm to generate cluster shears for an nfw (or other profile)- see [this class](https://github.com/LSSTDESC/CLMM/blob/c6cf806cd9b542f7286187ed0ecb6432f959a767/clmm/theory/parent_class.py) , especially `eval_convergence` and `eval_tangential_shear`

lbaumo:
here is an [example](https://github.com/LSSTDESC/CLMM/blob/c6cf806cd9b542f7286187ed0ecb6432f959a767/examples/demo_theory_functionality_oo.ipynb) of the cluster object

lbaumo:
we should probably set the maximum shear to .8 (though shear is definitely non-linear above g~.2) as large shears can come up in nfw profiles, and galsim cannot render very large shears properly

mr-superonion:
I am currently setting the maximum kappa to `0.6` (see [here](https://github.com/mr-superonion/descwl-shear-sims/blob/shear_obj/descwl_shear_sims/shear.py#L5)) to avoid the error message from Galsim when the distortions are too big.

PR #198, opened 2023-07-21
prepare the test for the implementation of shear from cluster

mr-superonion:
preparing to implement the shear object and get the shear from clmm (see issue: https://github.com/LSSTDESC/descwl-shear-sims/issues/197)


descwl_shear_sims/shear.py
```diff
@@ -0,0 +1,104 @@
+import clmm
+import galsim
+import numpy as np
+from astropy.coordinates import SkyCoord
```
beckermr, line 4: Let's make the astropy and clmm imports optional by importing them on the fly.

mr-superonion:
image
Trying to make simulation using shear from NFW halo.
But I am using the shear distortion from NFW halo to change the positions of background. Is it better to use the deflection angle?

esheldon:
I think using shear distortion is OK, but let's ask @rmjarvis to be sure

esheldon:
Is this failure in CI due to a change in the stack?  If so we may need to convert the type

mr-superonion:
It seems the failure is from the new version of the stack (The compiling succeed early yesterday...)
It seems to me that the problem comes from here:
https://github.com/mr-superonion/descwl-shear-sims/blob/b64f6d9de39f8eeef97086e749f8a121f3f2d29d/descwl_shear_sims/psfs/dmpsfs.py#L5
I do not know how to change the bases type..

esheldon:
expected list, got tuple

somewhere we are using a tuple instead of a list

esheldon:
@mr-superonion is simulating cluster shear and was wondering if he can just do this, for g1/g2 from the NFW profile he has

obj = obj.shear(g1=g1, g2=g2)

and get all the shears and sheared positions right, up to the level of accuracy we expect from galsim at shears up to tens of percent.  I thought it is fine but wanted to see if there are any gotchas.

rmjarvis:
That will shear the shapes, but not the positions.  You'd have to do the lensing deflection yourself to put the galaxy in the right place.  And that's not directly derivable from just g1,g2, so Position.shear is not the function you want.  You would need to go back to the raw deflection formula for NFW halos.

esheldon:
The positions move if the object is not at the origin.  Here the objects will be randomly placed on the image (see example image above) then the shear will be applied and the positions will get sheared.

rmjarvis:
I'm very confused then.  If you use `obj = obj.shear(g1,g2)`, that won't change the position. 
If you draw first and then shear the image, how do you do that?  The shear is not constant across the image, so there's not one shear that makes sense to apply.
Also, another thing you probably want to do if you plan to do cluster science with this kind of sim is to include the dilation/magnification.  `obj.lens()` can do the full calculation correctly given both shear and convergence.

esheldon:
Each object gets a shift and then an NFW shear.
Its really

obj = obj.shift( ).shear( )

We always do this because there is a slight bias in the recovered shear if we don't let the objects move, or  "shear the scene", as we say.

mr-superonion:
I update the requirements.txt and workflow and the error message with stackvana is solved

mr-superonion:
@lbaumo , Are there any functions in clmm to get the deflection angle fields for given halo profiles?

rmjarvis:
>> obj = obj.shift( ).shear( )

That wouldn't put things in the correct position for an NFW profile.  (Or indeed for any shear pattern other than a pure constant shear across the whole image.)  Indeed the sign of the effect isn't even right.  Central mass distributions push the positions out from the center of mass.  But your procedure will pull it back towards the center.

mr-superonion:
Thanks @rmjarvis , I will use the deflection angle from NFW to put the source galaxies to the correct position!!

mr-superonion:
@rmjarvis I realize that I do not understand why for constant shear field, the deflection of position is the same as the shear distortion?

rmjarvis:
Shear is the local derivative (Jacobian) of the deflection pattern.  If the shear is constant, then that derivative is the same at all points between the center and your galaxy.  So applying the shear to a shifted object correctly applies the same thing as the net deflection at the end of the shift.  But if the shear is variable, then the local derivative doesn't have to have anything to do with the global deflection.

esheldon:
I see thanks Mike.  Galsim can't know the shear is from an NFW.

I think this means you need to shear first then move the object @mr-superonion

mr-superonion:
OK. Thanks, I think now I understand. For a constant shear, 
1. the change in the relative position between any two galaxies is fully described by the Jacobian  (I was thinking about the absolute position; therefore, got confused.)
2. The Jacobian is the same in the space; therefore, deflecting galaxy centroid and shearing galaxy shape is commutative.

PR #200, opened 2023-10-18
Add Shear Object enabling Redshift-dependent Shear

mr-superonion:
@ismael-mendoza and I made this pull request for the redshift dependent shear simulation.
1. [This file](https://github.com/mr-superonion/descwl-shear-sims/blob/shear_obj_zshear/descwl_shear_sims/shear.py) contains the shear objects that is used to generate shear for different tests.
2. Unit test for this implementation can be found [here](https://github.com/mr-superonion/descwl-shear-sims/blob/shear_obj_zshear/shear_meas_tests/test_shear_meas_zshear_metacal.py).


mr-superonion:
@esheldon , @beckermr, Could you please review when you are free?

esheldon:
One general comment:  I would like to keep shear-measurement specific code out of this repo, for example the new code you added `bin/simulate_image.py`

descwl_shear_sims/sim.py
```diff
@@ -52,12 +52,13 @@ def make_sim(
     rng,
     galaxy_catalog,
     coadd_dim,
-    g1,
-    g2,
+    shear_obj,
```
esheldon, line 55: Can we make this somehow backwards compatible?

requirements.txt
```diff
@@ -1,8 +1,9 @@
-stackvana
+stackvana<=0.2023.28
```
esheldon, line 1: We should discuss this.  Were you having difficulty with new versions?

esheldon:
The code needs to pass flake8.

requirements.txt
mr-superonion: hmm, at some point it has an error but I forgot the details. Let me test this part...

descwl_shear_sims/sim.py
mr-superonion: Done, see [this unit test](https://github.com/mr-superonion/descwl-shear-sims/blob/shear_obj_zshear/shear_meas_tests/test_shear_meas_zshear_metacal_compatible.py) for backward compatibility!

mr-superonion:
Hello @esheldon , I checked that the latest `stackvana=0.2023.47` works. The newer version of mamba (`mamba>1.5.1`) seems to have some problems (see [here](https://github.com/mamba-org/mamba/issues/2994)). Therefore, I set `mamba==1.5.1` in the github workflow.

I think this PR is ready to merge.

descwl_shear_sims/sim.py
```diff
@@ -148,6 +151,11 @@ def make_sim(
 
     if se_dim is None:
         se_dim = get_se_dim(coadd_dim=coadd_dim, dither=dither, rotate=rotate)
+    if shear_obj is None:
+        assert isinstance(g1, float) & isinstance(g2, float), \
```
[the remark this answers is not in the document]
esheldon: another things a user might do is set g1/g2 from elements of numpy arrays which would also fail the test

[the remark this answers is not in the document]
mr-superonion: Thanks. revision made.

esheldon:
Thanks for the PR

Issue #188, opened 2022-10-17
Simulate HSC coadd image (calibrated)


mr-superonion:
Add configuration (magnitude zero point, seeing , noise level) for HSC coadded image and enable the code to provide multi-band simulations for HSC coadds.

PR #202, opened 2023-12-08
Enable users to use different survey setups (pixel_scale, psf_fwhm)

mr-superonion:
This PR enable users to use different survey setups. related to https://github.com/LSSTDESC/descwl-shear-sims/issues/188


mr-superonion:
@esheldon, @beckermr , can you please review this PR? It enable users to use different survey's pixel scales and magnitude zero points.

mr-superonion:
Hmm, I think I need more tests... now the code can make simulations for two observations (HSC: left, DES, right) with the same input galaxies, but they are not aligned well..
image

mr-superonion:
Let me add a test running the ML detection code and check the centroids have the same sky coords or not..

mr-superonion:
I think now the sky coordinate matched between the two sims:
image
I would say (again..) this is probably ready to merge...

esheldon (changes requested):
I like these changes generally.

However, it is really exposing limitations of the design choices we made.  For example the Layout objects are nice, and probably we should have used them from the beginning. Then we would not need to pass around `coadd_dims` everywhere.

It may be we eventually want to do a restructuring around survey objects and layout objects.  That would, for example, make it easy to simulate CCDs where coadd_dims is not relevant.  Also the survey would carry around pixel scale etc.

mr-superonion:
> make it easy to simulate CCDs where coadd_dims is not relevant.
Thanks!
Hmm, currently, we are using `coadd_dims` and `coadd_pixel_scale`, `sky_orgin` (not implemented, now it is fixed to (`WORLD_ORIGIN`)) passed to `Layout` object to define the galaxy distribution (taking flat-sky approximation). At the same time we are using `coadd_dims` and `pixel_scale` to define the wcs of the coadd image. 

Some possible changes are listed as follows:

- use `coadd_length` in units of arcmin to define the `Layout` object and remove `coadd_dims` from the input.
- change the `make_coadd_dm_wcs(layout, coadd_pixel_scale, coadd_dims)` to output a set of wcs for coadd pathches. 
- enable users to input wcs of single CCD.

esheldon:
Sorry, I was just discussing these as future possibilities, not for this PR

descwl_shear_sims/tests/test_pairs.py
```diff
@@ -1,10 +1,10 @@
 import pytest
 import numpy as np
-from ..galaxies import make_galaxy_catalog, DEFAULT_FIXED_GAL_CONFIG
-from ..psfs import make_fixed_psf
+from descwl_shear_sims.galaxies import make_galaxy_catalog, DEFAULT_FIXED_GAL_CONFIG
```
esheldon, line 3: Why was this changed?

esheldon (approved):

PR #206, opened 2024-02-27
Slightly Change the structure of Shear Object 

mr-superonion:
change the structure of shear_object so that I can use shear, convergence and deflection angle from halo e.g. see [here](https://github.com/mr-superonion/xlens/blob/86a283f7339c3ed08468fc50f57ab621ff44e28b/xlens/simulation/simulator/halo.py#L8).


esheldon:
Thanks @mr-superonion 

This looks OK to me, but it does seem like a separate change regarding density got put into this PR

esheldon (changes requested):
I think this is a good idea.

Can you please add some basic unit tests?

esheldon (approved):

PR #207, opened 2024-05-28
return dict of list of se_wcs

zhouconghao:
Corresponding [issue](https://github.com/LSSTDESC/descwl-shear-sims/issues/192)


esheldon (approved):
This breaking change is fine for all the codes I use.

PR #210, opened 2024-06-21
input galaxy index in truth_info

mr-superonion:
This PR make the following updates:
1. Add input galaxy index in truth_info;
2. Add unit test for the truth_info catalog;
3. Add one option in the simulation that do not add image noise to simulate noiseless images.

### DIFF
--- a/descwl_shear_sims/layout/layout.py
+++ b/descwl_shear_sims/layout/layout.py
@@ -45,7 +45,7 @@ class Layout(object):
             # in a square
             if (coadd_dim - 2*buff) < 2:
                 warnings.warn("dim - 2*buff <= 2, force it to 2.")
-                self.area = (2**pixel_scale/60)**2.  # [arcmin^2]
+                self.area = (2*pixel_scale/60)**2.  # [arcmin^2]
             else:
                 # [arcmin^2]
                 self.area = ((coadd_dim - 2*buff)*pixel_scale/60)**2


esheldon:
Yes it would be OK to make those dictionaries, keyed by band

descwl_shear_sims/galaxies.py
```diff
@@ -218,7 +219,7 @@ def get_objlist(self, *, survey):
             objlist.append(self._get_galaxy(flux))
             shifts.append(galsim.PositionD(sarray['dx'][i], sarray['dy'][i]))
 
-        return objlist, shifts, redshifts
+        return objlist, shifts, redshifts, indexes
```
esheldon, line 222: This is a breaking change.

Also the returned data is getting to be a lot.  Would we add more in the future?  Probably.

Maybe we should make this return a dict `{'objlist': objlist, 'shifts': shifts, ...}`

Since we are already making a breaking change this would be OK.

We should have a galaxies unit test that works with these and makes sure everything is constent.

descwl_shear_sims/tests/test_sim.py
```diff
@@ -634,6 +634,62 @@ def test_sim_des(psf_fwhm):
...
+        galaxy_catalog=galaxy_catalog,
+        coadd_dim=coadd_dim,
+        shear_obj=shear_obj,
+        psf=psf,
+        star_catalog=star_catalog,
+        calib_mag_zero=calib_mag_zero,
+        survey_name=survey_name,
+    )
+    assert "truth_info" in out.keys()
+    assert out["truth_info"].dtype.names == (
```
esheldon, line 688: should we also test that it is the same size as the indexes, and that the data are consistent?

esheldon (changes requested):

descwl_shear_sims/galaxies.py
esheldon: what do you think about this?

esheldon (changes requested):
Looks good. Can you please update the CHANGES.md?

setup.py
```diff
@@ -1,7 +1,7 @@
 import os
 from setuptools import setup, find_packages
 
-__version__ = None
+__version__ = ""
```
esheldon, line 4: why this change?

esheldon (approved):

PR #213, opened 2024-08-23
Add 1  in make_coadd_dm_wcs to account for the coordinate difference in dm and galsim

zhouconghao:
So we get the same origin when the coordinates are subtracted by 1 in make_dm_wcs.

### DIFF
--- a/descwl_shear_sims/wcs/dmwcs.py
+++ b/descwl_shear_sims/wcs/dmwcs.py
@@ -98,8 +98,8 @@ def make_coadd_dm_wcs(coadd_dim, pixel_scale=SCALE):
     coadd_origin = big_coadd_bbox.getCenter()
 
     gs_coadd_origin = galsim.PositionD(
-        x=coadd_origin.x,
-        y=coadd_origin.y,
+        x=coadd_origin.x + 1,
+        y=coadd_origin.y + 1,
     )
     coadd_wcs = make_dm_wcs(
         make_wcs(
@@ -132,8 +132,8 @@ def make_coadd_dm_wcs_simple(coadd_dim, pixel_scale=SCALE):
     coadd_origin = coadd_bbox.getCenter()
 
     gs_coadd_origin = galsim.PositionD(
-        x=coadd_origin.x,
-        y=coadd_origin.y,
+        x=coadd_origin.x + 1,
+        y=coadd_origin.y + 1,
     )
     coadd_wcs = make_dm_wcs(
         make_wcs(


beckermr (changes requested):
Let's add a test for this one to ensure it doesn't break later.

zhouconghao:
@beckermr Added a test for `make_simple_coadd_wcs`. The code is ready for review. Thanks!

descwl_shear_sims/tests/_wcs.py
```diff
@@ -12,7 +12,9 @@ def make_sim_wcs(dim):
 
     dims = [dim]*2
     cen = (np.array(dims)-1)/2
-    image_origin = galsim.PositionD(x=cen[1], y=cen[0])
+    # this is the center in 0 offset
+    # need to add 1 for the galsim convention
+    image_origin = galsim.PositionD(x=cen[1]+1, y=cen[0]+1)
```
beckermr, line 17: So this bug in our test sims ensured we missed the offset bug in the main code?

descwl_shear_sims/wcs/dmwcs.py
```diff
@@ -67,40 +69,48 @@ def make_dm_wcs(galsim_wcs):
...
+    # DM uses 0 offset, galsim uses FITS 1 offset
+    # we are converting from a dm position to a galsim position
+    # so we need to add 1
     gs_coadd_origin = galsim.PositionD(
-        x=coadd_origin.x,
-        y=coadd_origin.y,
+        x=coadd_origin.x + 1,
+        y=coadd_origin.y + 1,
     )
+    print("gs_coadd_origin", gs_coadd_origin)
```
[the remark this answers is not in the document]
zhouconghao: Fixed!

descwl_shear_sims/tests/_wcs.py
zhouconghao: I took a look at where this function is referenced. This function is used in `test_dmwcs.py`, `test_dmpsfs.py`, and `test_rand_psf.py`. 

In `test_dmwcs` in `test_dmwcs.py`, both `galsim` and DM WCSs use the wrong center since the DM WCS is produced from the galsim WCS. Therefore they agree with each other. 

I found this since in `make_coadd_dm_wcs` the center is calculated with the DM `bbox.getCenter()` method, which calculates the correct center that can be compared to the wrong `galsim` center. 

In `test_dmpsfs.py` and `test_rand_psf.py`, no DM WCS is used. The image is centered at the wrong image position but doesn't impact the test.

esheldon:
Is it true that when DM loads a WCS from a FITS header with 1 offset, it converts to zero offset?

beckermr:
TBH, I cannot tell from looking at the code. I am going to experiment with some images next week to figure out what it does in practice.

zhouconghao:
With @beckermr 's help, I read a DES image with DM. It does seem like DM shifts the offset. See here https://github.com/zhouconghao/work/blob/main/notebooks/test_dm_wcs_read_des.ipynb

PR #214, opened 2024-09-07
Enable option for using `making_coadd_dm_wcs_simple` in `make_sim`

zhouconghao:
Hi!

This PR adds an optional keyword argument in `make_sim` to enable switching between `make_coadd_dm_wcs` and `make_coadd_dm_wcs_simple`. The latter uses no padding and offset in making DM WCS and bbox, and the resulting world origins of SE WCS and coadd WCS are the same. 

I added a test to check if the world origins of SE and coadd are the same when using `make_coadd_dm_wcs_simple`. 

Thanks!

### DIFF
--- a/descwl_shear_sims/sim.py
+++ b/descwl_shear_sims/sim.py
@@ -136,11 +136,11 @@ def make_sim(
         the final coadd WCS and deteremines some properties of
         the single epoch images.
     simple_coadd_bbox: optional, bool. Default False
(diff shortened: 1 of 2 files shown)


zhouconghao:
The code is changed to force SE WCS to use WORLD ORIGIN as the sky origin if the `simple_coadd_wcs` option is used.

beckermr (approved):

PR #216, opened 2024-10-25
Override wcs in galaxy catalog if simple coadd bbox is used

zhouconghao:
Sorry didn't spot this in the last PR.

By default, `make_galaxy_catalog` calls `make_coadd_dm_wcs` to get a bbox with padding and attaches it to the layout. To make `simple_coadd_bbox` work with galaxy catalogs generated with `make_galaxy_catalog`, we need to override the bbox and WCS from the galaxy catalog layout.

### DIFF
--- a/descwl_shear_sims/sim.py
+++ b/descwl_shear_sims/sim.py
@@ -141,6 +141,8 @@ def make_sim(
         world origin. If set to False, the coadd bbox is embeded in a larger
         box bounded by (xoff, yoff, xoff+coadd_dim, yoff+coadd_dim) and
         the SE WCS has a different world origin compared to the coadd WCS.
+        This option overrides the wcs in galaxy catalog layout and requires
+        an input coadd_dim.
     draw_noise: optional, bool
         Whether draw image noise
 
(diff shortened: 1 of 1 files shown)


beckermr (commented):
We need @esheldon to look at this one. I am confused as to what is going on why catalogs come with WCS+bbox attributes.

esheldon:
The layout can have a wcs, and this was supposed to be optional 

https://github.com/LSSTDESC/descwl-shear-sims/blob/f33b18d0bcf7b8db4341e9df983c7183a06f401e/descwl_shear_sims/sim.py#L175

but it looks like it is always created for the layouts provided by this package.  I think we could probably not set the wcs in the layouts.   But let's ask @mr-superonion

esheldon:
One reason to set the wcs in the layout is to ensure consistency of the coordinate system

beckermr:
We probably should generate galaxy positions in the (u,v) tangent plane so it is independent of the WCS kind and location.

beckermr:
Ohhh I see what you mean. Yes you are right. Sorry for the noise!

esheldon:
I think we should probably refactor the code.

It would be better to generate catalogs in ra/dec, define a pointing (which could be random) and then create the projection/wcs around that point.

This would also facilitate the calibration simulations we have been discussing.

PR #217, opened 2024-10-27
Return prelensed coordinates, kappa, and gamma

zhouconghao:
Hi!

@mr-superonion and I are doing tests with cluster shear. It would be really helpful if we could get prelensed coordinates, κ and γ for each source. I implemented some code to return these quantities if the shear object is a `ShearHalo` from https://github.com/mr-superonion/xlens/blob/main/xlens/simulator/perturbation/halo.py . The code is a bit unwieldy now, but the changes do not impact other functionalities, I think. Let me know if any changes need to be made. Thanks!


zhouconghao:
Done! I guess we don't need to take `shift` to a tour so I deleted it in the return list of `ShearHalo`.

PR #219, opened 2024-11-14
Enable Simulation with User-defined WCS and BBOX in make_exp and galaxy.layout 

tae-h-shin:
Implemented an option to generate Roman-like simulation:
- added a new galaxy catalog object 
- created a function that generates a rubin-roman catalog with precalculated magnitudes and bulge fractions
- created new survey object, layout etc.

### DIFF
--- a/descwl_shear_sims/sim.py
+++ b/descwl_shear_sims/sim.py
@@ -9,7 +9,7 @@ from lsst.afw.cameraGeom.testUtils import DetectorWrapper
 from .lsst_bits import get_flagval
 from .saturation import saturate_image_and_mask, BAND_SAT_VALS
 from .surveys import get_survey, rescale_wldeblend_exp, DEFAULT_SURVEY_BANDS
-from .constants import SCALE, WORLD_ORIGIN, ZERO_POINT
+from .constants import SCALE, WORLD_ORIGIN, ZERO_POINT, SIM_INCLUSION_PADDING
 from .artifacts import add_bleed, get_max_mag_with_bleed
 from .masking import (
     get_bmask_and_set_image,
@@ -660,9 +660,11 @@ def _draw_objects(
         prelensed_image_pos = wcs.toImage(prelensed_world_pos)
 
         if (
-            (image.bounds.xmin - 80) < image_pos.x < (image.bounds.xmax + 80)
+            (image.bounds.xmin - SIM_INCLUSION_PADDING) <
+            image_pos.x < (image.bounds.xmax + SIM_INCLUSION_PADDING)
         ) and (
-            (image.bounds.ymin - 80) < image_pos.y < (image.bounds.ymax + 80)
+            (image.bounds.ymin - SIM_INCLUSION_PADDING)
+            < image_pos.y < (image.bounds.ymax + SIM_INCLUSION_PADDING)
         ):
             local_wcs = wcs.local(image_pos=image_pos)
             convolved_object = get_convolved_object(obj, psf, image_pos)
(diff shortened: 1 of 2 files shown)


esheldon:
Here are some general comments, I'll put more specific comments in a proper review

- let's keep functions defined in their own modules rather than in `__init__.py`.  Thus `make_galaxy_catalog` should go into some module.
- We use `pep8` style for module names, lowercase with underscores rather than CamelCase

esheldon:
A general discussion topic:  does it make sense to have Rubin sims in this repo?  Would it be better to have a separate `rubin-lsst` package?

esheldon:
Sorry, yes, I meant Roman.

beckermr (dismissed):
My $0.02 is that it does NOT make sense to support Roman sims here. 
The original purpose of this package was to create simple, easy-to-debug, and well-validated simulations to do testing of shear codes for the DESC in the weak shear regime. 

This package is not meant for:

- work to calibrate shear estimators for specific data releases or analyses
- work on surveys besides LSST
- work on shear estimation in other regimes like clusters

As we add all of these features, we make the code harder to debug and harder to validate.

beckermr:
If we plan to add all of these features, we might as well be using the galsim config parser. 

Again this:

> The original purpose of this package was to create simple, easy-to-debug, and well-validated simulations to do testing of shear codes for the DESC in the weak shear regime.

esheldon:
I think new catalogs do not necessarily need to be part of this package.  The `make_sim` function takes catalogs as an argument, so the user can define their own, in their own package that uses `descwl-shear-sims`, without making it part of this package.

Generally that is the design, to let the user send the objects.  I have a number of my own separate codes that use different catalogs.

However, there are some limitations to that. For example it doesn't currently allow sending an object to describe the instrument scale etc.

Also, how we would deal with realistic survey layouts is "kludgy" with the current setup.

I think we have two choices here:

1. Refactor the package so that all aspects of the simulation are sent as arguments.  This would allow the user to define their own objects without increasing the size of this code.
2. Create a new, more general package that has the above design elements:  all objects can be created by the user, but with defaults for LSST.

descwl_shear_sims/sim.py
```diff
@@ -174,30 +171,41 @@ def make_sim(
...
+        else:
+            world_origin = WORLD_ORIGIN
+        coadd_wcs, coadd_bbox = make_coadd_dm_wcs_simple(
+            coadd_dim,
+            pixel_scale=pixel_scale,
+            world_origin=world_origin,
+        )
+    else:
+        if (
+            hasattr(galaxy_catalog.layout, "wcs")
```
esheldon, line 195: I don't like the wcs "hiding" in the galaxy catalog.

Maybe we should have an optional argument to `make_sim` that can hold coadd information.

e.g.
```python
coadd_info = {'coadd_wcs': ..., 'coadd_bbox': ...}\
data = make_sim(..., coadd_info=coadd_info)
```

esheldon: Note in layout.py the wcs and bbox are not actually used.

descwl_shear_sims/sim.py
```diff
@@ -658,17 +659,21 @@ def _draw_objects(
         )
         prelensed_image_pos = wcs.toImage(prelensed_world_pos)
 
-        local_wcs = wcs.local(image_pos=image_pos)
-
-        convolved_object = get_convolved_object(obj, psf, image_pos)
+        if (
+            (image.bounds.xmin - 80) < image_pos.x < (image.bounds.xmax + 80)
```
esheldon, line 663: How did you determine 80?  Should this be in constants.py with an explanation?

descwl_shear_sims/sim.py
esheldon: What about this?

esheldon (approved):

PR #227, opened 2025-05-20
Enable probability weight for input catalog

mr-superonion:
This PR enable users to reweight the input galaxy catalog to match to the real data.

### DIFF
--- a/.github/workflows/shear_meas_tests.yml
+++ b/.github/workflows/shear_meas_tests.yml
@@ -12,7 +12,7 @@ jobs:
     name: shear-meas-tests
     strategy:
       matrix:
-        pyver: ["3.11"]
+        pyver: ["3.12"]
 
     runs-on: "ubuntu-latest"
 


esheldon (approved):

descwl_shear_sims/galaxies.py
```diff
@@ -696,6 +696,14 @@ def __init__(
             select_lower_limit=select_lower_limit,
             select_upper_limit=select_upper_limit,
         )
+        ngal = len(self._wldeblend_cat)
+        if "prob" in self._wldeblend_cat.dtype.names and ngal > 0:
```
esheldon, line 700: This field is not in the standard file OneSqDeg.fits

This creates a situation where the code would silently do something different based on the file data, but it is still internally assuming OneSqDeg.fits is still the name, which is confusing and could cause bugs.

I think maybe this should not be part of this class. I think it makes more sense for you to create your own galaxy class that expects a different input file.

PR #240, opened 2026-04-15
mamba not found, use conda

esheldon:
This won't run till after merged


esheldon:
This is for shear meas. tests that don't run unti after we merge the PR....

